PR #39, opened 2021-04-13
Improve handling Docker

cameronwp:
After deploying MMGIS with Docker, here are a few little changes that should make it easier for folks to use Docker in the future.

1. Reorganizes the Dockerfile to take advantage of layer caching. With the new Dockerfile, you can rebuild the image when source code changes without reinstalling all dependencies. FYI, if you ever want to force Docker to rebuild all layers and reinstall the GDAL and PHP dependencies, you can use the `--no-cache` flag
2. Adds a docker-compose file that's ready to use. Includes instructions.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,9 @@
 FROM node:12
 
+# Install PHP and GDAL
+RUN apt-get update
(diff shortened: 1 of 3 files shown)


tariqksoliman (approved):
These changes look great. Thanks for the help!

PR #45, opened 2021-04-21
Ignore .env file and git directories in Docker

jtroberts:


tariqksoliman:
Good catch! Can you push to development though. Master only changes for releases.

tariqksoliman (approved):

PR #46, opened 2021-04-21
Auxiliary bulk_tiles tool to generate tile layers from tiffs in bulk.

jtroberts:
Also updated rasterstotiles and rastertolegend to work better with new
tool (mostly friendlier file path handling) and be more compliant with
Python3 and Flake8.

### DIFF
--- a/auxiliary/rasterstotiles/rasterstotiles.py
+++ b/auxiliary/rasterstotiles/rasterstotiles.py
@@ -1,12 +1,12 @@
-import sys
 import subprocess
-from osgeo import gdal,ogr,osr
+from osgeo import gdal, osr
 from argparse import ArgumentParser, ArgumentDefaultsHelpFormatter
 
-#Windows
-GDAL2TILE_PATH = 'C:\Python27\Lib\site-packages\osgeo\scripts\gdal2tiles.py'
-#Linux
-#GDAL2TILE_PATH = '/usr/local/bin/gdal2tiles.py'
+# Windows
+# GDAL2TILE_PATH = 'C:\Python27\Lib\site-packages\osgeo\scripts\gdal2tiles.py'
+# Linux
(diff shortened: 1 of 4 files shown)


auxiliary/rasterstotiles/rasterstotiles.py
```diff
@@ -1,12 +1,12 @@
...
 from argparse import ArgumentParser, ArgumentDefaultsHelpFormatter
 
-#Windows
-GDAL2TILE_PATH = 'C:\Python27\Lib\site-packages\osgeo\scripts\gdal2tiles.py'
-#Linux
-#GDAL2TILE_PATH = '/usr/local/bin/gdal2tiles.py'
+# Windows
+# GDAL2TILE_PATH = 'C:\Python27\Lib\site-packages\osgeo\scripts\gdal2tiles.py'
+# Linux
+GDAL2TILE_PATH = '/usr/local/bin/gdal2tiles.py'
```
tariqksoliman, line 8: I'm a little uncomfortable with this change. I guess it's not generic enough either way.

tariqksoliman (approved):

auxiliary/rasterstotiles/rasterstotiles.py
jtroberts: Yeah, I think there's a generic cross platform way that might be worth looking into. For Linux, I would expect users to already have the tool in their $PATH and it shouldn't need to be specified. Not sure how Windows handles things.

tariqksoliman: Go ahead and merge anyway.

fcalef: Agreed. We need to find a better way to address multiplatform Python programs that reference GDAL, but let's merge it anyways since it doesn't conflict with other programs. Also, it's a simple fix to change the GDAL2TILE_PATH variable.

PR #50, opened 2021-04-28
Allow dynamic updates of vector layers via function call

ac-61:
This allows dynamic updates of vector layers by exposing functions in `src/essence/API/API.js`. The two exposed functions can be called using `window.API`.

The following functions are available:
-  `window.API.clearVectorLayer`
- `window.API.updateVectorLayer`

This adds a new `controlled` parameter to the configuration for vector layers. If the `controlled` option is checked, the URL path to a geojson is not required.
Screen Shot 2021-04-27 at 7 11 05 PM


tariqksoliman (commented):
Nice, looks great! Feel free to merge.

PR #63, opened 2021-05-24
Check for WMTS layer to prevent globe from wrongly flipping tiles

jtroberts:


tariqksoliman (approved):
Simple enough though it won't work well for layers with mixed tile formats. The globe is being pulled out into a standalone npm package and this is a task is it too.

PR #64, opened 2021-06-03
Updated scripts in private/api to be compatible with Python3 and GDAL3

jtroberts:
Changes are backward compatible. 

Note that other remaining Python files found in localization/ haven't been updated. I'm not sure how to test these since they appear to be MSL mission specific. I could do a bulk update of print statements though.

### DIFF
--- a/private/api/getprofile.php
+++ b/private/api/getprofile.php
@@ -7,5 +7,5 @@
   $steps = escapeshellarg(urldecode($argv[6]));
   $axes = escapeshellarg(urldecode($argv[7]));
 
-  echo json_encode(exec("python3 private/api/2ptsToProfile.py $path $lat1 $lon1 $lat2 $lon2 $steps $axes 1 2>&1" ));
+  echo json_encode(exec("python private/api/2ptsToProfile.py $path $lat1 $lon1 $lat2 $lon2 $steps $axes 1 2>&1" ));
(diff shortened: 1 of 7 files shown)


private/api/getprofile.php
```diff
@@ -7,5 +7,5 @@
   $steps = escapeshellarg(urldecode($argv[6]));
   $axes = escapeshellarg(urldecode($argv[7]));
 
-  echo json_encode(exec("python private/api/2ptsToProfile.py $path $lat1 $lon1 $lat2 $lon2 $steps $axes 1 2>&1" ));
+  echo json_encode(exec("python3 private/api/2ptsToProfile.py $path $lat1 $lon1 $lat2 $lon2 $steps $axes 1 2>&1" ));
```
tariqksoliman, line 10: Do we know the command will be `python3`? Does the Dockerfile's python from `RUN apt-get install -y php7.0-cli php-db php-sqlite3 gdal-bin libgdal-dev python-pip python-gdal` support it?

jtroberts: I just updated the Dockerfile so that "python" refers to Python3 and ensured that the GDAL bindings are included in Python3. Also changed "python3" calls back to just "python" to allow any older systems that aren't using Python3 or Docker to still run the PHP scripts.

PR #67, opened 2021-06-16
Add helper functions to API for finding map features

ac-61:
This adds helper functions to the API to allow users to find map features such as the current features contained the current map view and the selected feature. This also adds 3 types of listeners (`onPan` , `onZoom`, `onClick`), which allows users to  connect these events with custom functions.

All new functions are documented in `docs/pages/markdowns/JavaScript_API.md`.

This also renames the global variable for the API from `window.API` to `window.mmgisAPI`.


tariqksoliman (approved):
Looks good to me!

PR #70, opened 2021-06-29
Clean up PHP deps in Dockerfile and README + better GDAL python imports

jtroberts:
Removed unnecessary PHP dependencies from Dockerfile and updated README. Also updated Python GDAL imports in API scripts for better compatibility with some Python GDAL installations.


private/api/2ptsToProfile.py
```diff
@@ -8,10 +8,10 @@
 # Setting axes to "xyz" will return image coordinates instead of lat lons
 
 import sys
-import gdal
-import osr
 import math
-from gdalconst import *
+from osgeo import gdal
```
tariqksoliman, line 12: Can you verify that `from osgeo import gdal` works from the docker container.

tariqksoliman (changes requested):

private/api/2ptsToProfile.py
tariqksoliman: Accidentally set it as requested changes but it's just a comment and you can merge once verified.

jtroberts: I built a Docker container with the latest changes and the scripts worked as expected.

PR #73, opened 2021-07-01
Add onLoaded function to MMGIS API

ac-61:
This adds an `onLoaded` function to the MMGIS API, which calls the input function once MMGIS has finished loading.

An example of how to call the `onLoaded` function:

```javascript
window.mmgisAPI.onLoaded(() => { 
    // Add listener with MMGIS API now that MMGIS has finished loading
    function listener() {
        const featuresContained = window.mmgisAPI.featuresContained()
        console.log('featuresContained', featuresContained)
        const activeFeature = window.mmgisAPI.getActiveFeature()
        console.log('activeFeature', activeFeature)
    }
    window.mmgisAPI.addEventListener('onPan', listener)
})
```

### DIFF
--- a/src/essence/mmgisAPI/mmgisAPI.js
+++ b/src/essence/mmgisAPI/mmgisAPI.js
@@ -137,7 +137,7 @@ var mmgisAPI_ = {
     writeCoordinateURL: function() {
         return QueryURL.writeCoordinateURL(false);
     },
-    onLoadCallback: false,
+    onLoadCallback: null,
     onLoaded: function(onLoadCallback) {
         mmgisAPI_.onLoadCallback = onLoadCallback
     },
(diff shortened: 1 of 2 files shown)


src/essence/mmgisAPI/mmgisAPI.js
```diff
@@ -136,6 +137,38 @@ var mmgisAPI_ = {
     writeCoordinateURL: function() {
         return QueryURL.writeCoordinateURL(false);
     },
+    isLoaded: false,
+    onLoaded: function(functionReference, timeout = 60000) {
```
tariqksoliman, line 141: This is more complicated than it needs to be. I think you can just do:
```javascript
fina: function() {
    //...
    if( typeof mmgisAPI_.onLoadCallback === 'function' ) mmgisAPI_.onLoadCallback()
},
onLoaded: function(onLoadCallback) {
    mmgisAPI_.onLoadCallback = onLoadCallback
}
```

tariqksoliman (changes requested):

tariqksoliman (approved):
Looks good now!

PR #75, opened 2021-07-08
Added Measurement Feature: Unit Displayed

Ricecrackerz:
Users are now able to select from a dropdown bar in the measurement tool to select which unit they want displayed.
-Added display units Meters & Kilometers
-Added a dropdown bar in the measurement tool to select between M & KM
-KM Float value is fixed to 2
-The unit is displayed in both the cursor hover and when actually creating the points
-When the unit is changed, the measurement is reset
-Units also work when the mode is set to "continuous"
-The branch outside of MeasureTool.js has outdated code.

@jtroberts

### DIFF
--- a/src/essence/Tools/Measure/MeasureTool.js
+++ b/src/essence/Tools/Measure/MeasureTool.js
@@ -537,7 +537,7 @@ function makeMeasureToolLayer() {
                     )
                 }else if(distDisplayUnit == "kilometers"){
                     temp.bindTooltip(
-                        '' + (roundedTotalDist/1000) + 'km ' + distAzimuth + '&deg;',
+                        '' + (roundedTotalDist/1000).toFixed(2) + 'km ' + distAzimuth + '&deg;',
                         {
                             permanent: true,
                             direction: 'right',
@@ -802,7 +802,7 @@ function makeGhostLine(lng, lat) {
             )
         }else if (distDisplayUnit == "kilometers"){
             CursorInfo.update(
-                `${roundedTotalDist/1000}km ${
+                `${(roundedTotalDist/1000).toFixed(2)}km ${
                     mode === 'continuous' ? `(+${roundedDist/1000}km)` : ''
                 } ${distAzimuth}&deg;`,
                 null,
--- a/src/essence/Tools/Measure/MeasureTool.js
+++ b/src/essence/Tools/Measure/MeasureTool.js
@@ -44,10 +44,10 @@ const Measure = () => {
             .on('click', MeasureTool.clickMap)
             .on('mousemove', MeasureTool.moveMap)
             .on('mouseout', MeasureTool.mouseOutMap)
-        // Globe_.litho
-        //     .getContainer()
-        //     .addEventListener('click', MeasureTool.clickGlobe, false)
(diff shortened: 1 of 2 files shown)


src/essence/Tools/Measure/MeasureTool.js
```diff
@@ -764,13 +792,23 @@ function makeGhostLine(lng, lat) {
...
+                `${roundedTotalDist}m ${
+                    mode === 'continuous' ? `(+${roundedDist}m)` : ''
+                } ${distAzimuth}&deg;`,
+                null,
+                false
+            )
+        }else if (distDisplayUnit == "kilometers"){
+            CursorInfo.update(
+                `${(roundedTotalDist/1000).toFixed(2)}km ${
+                    mode === 'continuous' ? `(+${roundedDist/1000}km)` : ''
```
jtroberts, line 806: I think you also want to have `toFixed(2)` here for consistency.

src/essence/Tools/Measure/MeasureTool.js
```diff
@@ -43,10 +44,10 @@ const Measure = () => {
...
             .on('mousemove', MeasureTool.moveMap)
             .on('mouseout', MeasureTool.mouseOutMap)
-        Globe_.litho
-            .getContainer()
-            .addEventListener('click', MeasureTool.clickGlobe, false)
-            .addEventListener('mousemove', MeasureTool.moveGlobe, false)
+        // Globe_.litho
+        //     .getContainer()
+        //     .addEventListener('click', MeasureTool.clickGlobe, false)
+        //     .addEventListener('mousemove', MeasureTool.moveGlobe, false)
```
jtroberts, line 51: There is a merge conflict that needs to be resolved here.
```
        const globeCont = Globe_.litho.getContainer()
        globeCont.addEventListener('click', MeasureTool.clickGlobe, false)
        globeCont.addEventListener('mousemove', MeasureTool.moveGlobe, false)
```

tariqksoliman: Yep, the (now) existing `const globeCont = Globe_.litho...` should be good.

src/essence/Tools/Measure/MeasureTool.js
Ricecrackerz: Oh I totally missed that, will fix that.

Ricecrackerz:
I attempted to push that small change of adding a floating point to the continuous KM measurement. The code might still be outdated with the globe litho stuff.

src/essence/Tools/Measure/MeasureTool.js
```diff
@@ -1,858 +1,897 @@
...
+        // code to run on component mount
+        Map_.map
+            .on('click', MeasureTool.clickMap)
+            .on('mousemove', MeasureTool.moveMap)
+            .on('mouseout', MeasureTool.mouseOutMap)
+
+        const globeCont = Globe_.litho.getContainer()
+        globeCont.addEventListener('click', MeasureTool.clickGlobe, false)
+        globeCont.addEventListener('mousemove', MeasureTool.moveGlobe, false)
+        development
```
jtroberts, line 51: This line shouldn't be in here.

It also looks like the last commit removed a bunch of white spaces.

jtroberts (approved):

PR #82, opened 2021-07-26
Allow draw tool features to behave similar to layer features

ac-61:
This updates the features created via the draw tool to behave similar to layer features when the draw tool panel is hidden. All of the normal draw tool functions return when the draw tool panel is visible.

Once the draw tool panel is hidden, draw tool features that are currently visible:
- Can be selected by clicking the feature and the selected feature information is displayed in the InfoTools and in the description of the top toolbar 
- Can be deselected by clicking anywhere on the map or another feature
- If a feature is selected, it will be returned as the active feature with `window.mmgisAPI. getActiveFeature()`

Draw tool layers that were visible at least once will:
- Appear as one of the objects returned by `window.mmgisAPI.featuresContained()` if any of the layer's features are in the current map bounds
- Appear as one of the objects returned by `window.mmgisAPI.getVisibleLayers()`

If a draw tool feature is selected and then the draw tool panel is made visible, the selected draw tool feature will:
- Maintain its selected highlight color after hovering over the selected draw tool feature on the map
- Maintain its selected highlight color after hovering over the draw tool layer name in the draw tool panel (i.e. after all features in a draw tool layer are highlighted when hovering over a draw tool layer name)
- Be deselected by clicking anywhere on the map that is not a draw tool feature

Draw tool layers have unique names that starts with `DrawTool_` and an appended number. The user provided name for a given draw tool layer is saved under the feature object as the property `layerName`.

### DIFF
--- a/src/essence/Basics/Layers_/Layers_.js
+++ b/src/essence/Basics/Layers_/Layers_.js
@@ -4,6 +4,7 @@ import Description from '../../Ancillary/Description'
 import Search from '../../Ancillary/Search'
 import ToolController_ from '../../Basics/ToolController_/ToolController_'
 import $ from 'jquery'
+import * as d3 from 'd3'
 
 var L_ = {
     url: window.location.href,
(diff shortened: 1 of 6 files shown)


tariqksoliman (approved):
LGTM.
I made just a few changes (though the formatting makes it look like more):

1. Wrapped the `e.on(...)` callback functions in a closure
2. Used `Map_.activeLayer` instead of `infoTool.currentLayer` (tools should never rely on each other)

You can merge changes but I did also notice a small issue with opening the DrawTool when a feature from one of its layers is already highlighted.

PR #84, opened 2021-07-27
New Feature to add symbols and dynamic colors

Ricecrackerz:
Implemented:
-A dropdown bar to choose a shape for a layer (a point on the map) in the config page
-Shapes such as: Triangle, Square, Circle, Star, Hexagon, and Dinosaur
-Changed the selected stroke color to white
-Edited the props:color code to let symbols have a fill color based on some database
-Added a triangle shape to the Legends.js file to display a Triangle.


src/essence/Basics/Map_/Map_.js
```diff
@@ -1004,7 +1004,67 @@ function makeLayer(layerObj) {
                         latlong,
                         leafletLayerObject.style
                     ).setRadius(layerObj.radius)
-
+                    
+                    
+                    if(true){
```
jtroberts, line 1009: I think you can get rid of this if statement and the console log message below.

jtroberts (approved):

tariqksoliman (changes requested):
Nice work!

I think we really need a "none" type (which is the default (not "circle")) that uses the previous `L.circleMarker` class. Just so that everything existing works the same way.

You should also use the rest of the style properties in the svgs:

- stroke = Stroke Color
- stroke-width = Stroke Weight
- opacity = Fill Opacity
- r = Radius (a bit trickier than the rest too since you need to change the width and height style props as well)

And, as much as it pains me to say, the dinosaurs are going to have to go extinct.

tariqksoliman (approved):
LGTM

PR #85, opened 2021-07-27
Clear the active layer when clicking on map area with no features

ac-61:
This should fix the "small issue" mentioned in https://github.com/NASA-AMMOS/MMGIS/pull/82#pullrequestreview-716187777.


tariqksoliman (approved):
Ah, so it was my fault!

ac-61:
@tariqksoliman And mine! I didn't realize we had to clear this value out. 😁

PR #86, opened 2021-07-28
Bug fixes for DrawTool layers

ac-61:
This fixes 2 bugs:
1. The highlight for DrawTool points was not cleared when clicking on map areas where there are no features
2. Clicking DrawTool text annotations did not trigger the mmgisAPI `onClick` event listener


tariqksoliman (approved):
I had no idea optional chaining was a thing!

PR #89, opened 2021-08-04
Allow for custom LayersTool width for missions with long layer names

jtroberts:


tariqksoliman (approved):
This is a cool solution.

PR #97, opened 2021-08-19
Add Isochrone tool

frasercl:
Adds a new Isochrone tool for analyzing and planning traverses. Features:
- Render isochrones as raster layers showing color-ramped accumulated cost out to a set maximum
- Hover over isochrones to show least-cost paths and minimum cost from start point to hovered point
- Separated cost model interface allows easy extension and switching models on-the-fly
- Add, remove, and separately edit multiple isochrones
- Configurable color ramps, color steps, opacity, resolution, data sources

### DIFF
--- a/src/essence/Tools/Isochrone/IsochroneTool_Algorithm.js
+++ b/src/essence/Tools/Isochrone/IsochroneTool_Algorithm.js
@@ -1,0 +1,232 @@
+
+import Map_ from "../../Basics/Map_/Map_";
+
+/*
+Runs Dijkstra's Algorithm on a given subsection of the map
+
+Dijkstra's Algorithm operates on a weighted graph, and finds the lowest-weight path between a given
+starting vertex and all other vertices. 
+
+In this implementation, values in a 2D array (representing pixels of a raster) are treated as
+vertices of the graph and are considered to be implicitly connected to their immediate neighbors.
+A cost function is run between connected pixels to determine edge weights. The result of the
+algorithm is returned as two arrays: a cost array which stores the minimum cost of reaching every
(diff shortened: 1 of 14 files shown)


src/essence/Tools/Isochrone/IsochroneTool_Algorithm.js
```diff
@@ -0,0 +1,232 @@
...
+    [ 0,  1,  0,  9,  0],
+    [16, 15, 13, 11, 10],
+    [ 0, 14,  0, 12,  0]
+];
+
+/**
+ * Moves from a pixel to its 8 immediate neighbors.
+ * A purely distance-based model will produce an octagon with these moves.
+ */
+const queenMoves = [
```
tariqksoliman, line 141: So she's restricting her movement to not make the king feel inferior?

tariqksoliman (approved):
LGTM and thanks for pulling this in!

src/essence/Tools/Isochrone/IsochroneTool_Algorithm.js
frasercl: Don't know a lot about chess haha - I believe that's a term I saw somewhere else, and I didn't think about it too much. Would probably more correctly be called `kingMoves`.

tariqksoliman: `queen` is fine; the comment and 8 elements explain it pretty easily.

frasercl: And those moves can compose to queen moves, so it's not as if it's completely inaccurate.

PR #98, opened 2021-08-26
Polar map compatibility fixes for viewshed tool

frasercl:
Changes to viewshed tool for additional projection & configuration support, including:
- Support for negative tile coordinates
  - e.g. check map boundaries via `L.Map.getPixelWorldBounds` when flipping Y tile coordinates
- Replaced some LatLng-based calculations in `ViewshedTool_Manager.updateDesiredTiles` with calculations in map pixel space
  - in particular, factored out `L.LatLngBounds`, which are universally unsafe on polar maps
- Rounded all references to `L.Map.getZoom`, to support non-integer zoom levels (Leaflet does this internally as well)
- Replaced one Lithosphere projection function that was giving bad results with a Leaflet equivalent (in `ViewshedTool_Manager.locateSource`)

### DIFF
--- a/src/essence/Tools/Viewshed/ViewshedTool.js
+++ b/src/essence/Tools/Viewshed/ViewshedTool.js
@@ -1157,6 +1157,14 @@ let ViewshedTool = {
                         dlc[z] = dlc[z] || {}
                         dlc[z][Math.floor(x)] = dlc[z][Math.floor(x)] || {}
 
+                        const tileRow =
+                            (y - Math.floor(data.outputTopLeftTile.y)
+                            - (Math.abs(data.outputTopLeftTile.y) % 1) * 2) * res
+
+                        const tileCol =
+                            (x - Math.floor(data.outputTopLeftTile.x)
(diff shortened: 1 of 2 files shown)


tariqksoliman (approved):
LGTM! Thanks for adding this change! Also what was the reasoning for removing the cleanupSeams functions (at L72 and L86). It might not be ideal and is hard to see but it does somewhat minimize the artifacts from the tile seams.

frasercl:
> LGTM! Thanks for adding this change! Also what was the reasoning for removing the cleanupSeams functions (at L72 and L86). It might not be ideal and is hard to see but it does somewhat minimize the artifacts from the tile seams.

`cleanupSeams` had an immediate `return` statement that blocked the rest of the function (L495) - I removed the return statement along with calls to the function to preserve the original behavior. That change isn't necessary to the fixes I made and can be safely reverted, though reverting only the removal of function calls will introduce the code in `cleanupSeams` to the data pipeline when it wasn't there before.

tariqksoliman:
Ah, you're completely right. Looks like I deprecated it in favor of interpolateSeams. Feel free to merge!

PR #104, opened 2021-09-14
Minor updates to auxiliary and other helper scripts

jtroberts:
- Allow marker radius to be set by feature radius property
- Option to calculate distance between points with Vincenty's formulae
- Allow for optional help page route
- Use percentages for sample color table
- Fixed epsg typo in rasterstotiles
- Allow processing without colormap and include prefix in output filename using bulk_tiles
- Ignore pycache, ds_store, and extra data dir


tariqksoliman (approved):
LGTM

PR #116, opened 2021-11-24
#115 MeasureTool - Show topography on all line segments

### LINKED ISSUE #115: MeasureTool - Show topography on all line segments
> Currently the "Continuous" mode in MMGIS's Measure tool allows the user to create multiple line segments continuously, but the elevation profile displayed is only of the last segment that was drawn. Instead, we would like for the elevation profile to display the elevation profile for the entire drawn line, such that each new line segment just adds to the existing elevation profile.

tariqksoliman:
Closes #115 

measuretool_continuous_profile


tariqksoliman:
- [ ] Globe: Show red measure lines
- [ ] Hover tooltip remove index number
- [ ] Hover tooltip show topological distance (3d calculated each sample point)
- [ ] Hover tooltip anchors top/bottom
- [ ] Continuous colorful

tariqksoliman:
Example of "Continuous Color" mode:
measure_cont_color

Issue #117, opened 2021-12-06
Support longer names in legend tool


jtroberts:
I'm working with one mission that has long names for some of the legend items that are overrunning the available width of the legend tool.

PR #118, opened 2021-12-06
#117 Increased width of legend tool and allow line breaks when needed

jtroberts:


tariqksoliman (approved):
LGTM!
(and do remember, in case of potential larger legend changes, that we have thoughts on making the legend independent of tools)

PR #155, opened 2022-03-02
Fix minor bugs

ac-61:
This adds a fix suggested by @tariqksoliman to fix a bug that would cause a crash if an item's display name is a number. This also fixes a bug where clicking on the map with no features did not clear the info tool and description box.


tariqksoliman (approved):
LGTM and thanks for finding these and getting them in!

PR #156, opened 2022-03-03
Add LineString functions

ac-61:
This adds two new functions to perform operations on vector layers containing features made up of `LineString`.

## trimLineString
This function can be used to trim a specified number of vertices on a specified layer containing GeoJson LineString features. The function can trim from either the beginning or end.

This function makes the following assumptions:
- The user calling the method knows exactly how many vertices they want to trim and the new start/end time
- The features are ordered based on time in ascending order
- All of the features in the layer must have `LineString` as the type
- If trimming from the beginning of the layer, the time in the `time` parameter must be after the start time of the first feature in the layer
- If trimming from the end of the layer, the time in the `time` parameter must be before the end time of the last feature in the layer

## appendLineString
This function appends input data with a GeoJson Feature that contains a LineString. The LineString vertices from the input data is appended as vertices to the last Feature in the layer. The new end time is updated using the information from the input data.

This function makes the following assumptions:
- There is at least one feature in the layer to append to
- The input data is a single GeoJson feature. Below is an example of the expected input data format:
    ```
     {
        'type':'Feature',
        'properties':{
            'name': 2,
            'Length_m': 0,
            'COLOR': 0,
            'route': 0,
            'start_time': '2021-12-01T15:10:00.000Z',
            'end_time': '2021-12-01T15:20:00.000Z'
        },
        'geometry':{
                'type':'LineString',
                'coordinates':[
                    [145.862136,-73.208439],
                    [135.063782,-71.898251],
                    [130.828697,-75.540527],
                    [122.767247,-72.658683],
                    [120.133499,-75.018059]
                ]
        }
    }
    ```

### DIFF
--- a/src/essence/Basics/Layers_/Layers_.js
+++ b/src/essence/Basics/Layers_/Layers_.js
@@ -2108,4 +2108,3 @@ function parseConfig(configData, urlOnLayers) {
 }
 
 export default L_
-window.L_ = L_
--- a/src/essence/Basics/Layers_/Layers_.js
+++ b/src/essence/Basics/Layers_/Layers_.js
@@ -1584,18 +1584,34 @@ var L_ = {
             )
         }
     },
-    trimLineString: function (layerName, startTimeData, endTimeData, trimN, trimDirection) {
-        // Validate input parameter
-        if (!startTimeData || !endTimeData) {
+    trimLineString: function (layerName, time, timeProp, trimN, startOrEnd) {
+        // Validate input parameters
+        if (!time) {
             console.warn(
                 'Warning: Unable to trim the LineString in vector layer `' +
                     layerName +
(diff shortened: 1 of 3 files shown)


src/essence/Basics/Layers_/Layers_.js
```diff
@@ -1812,3 +2108,4 @@ function parseConfig(configData, urlOnLayers) {
 }
 
 export default L_
+window.L_ = L_
```
tariqksoliman, line 2111: Is this needed and where?

ac-61: Fixed 118971b12711ba13af5483224bc2295e9d8d0cbe!

tariqksoliman (approved):
Looks awesome!

PR #158, opened 2022-03-14
Update the globe when vector layers are modified

ac-61:
This updates the vector layer displayed on the globe whenever it is modified using any of the functions in the javascript API.

### DIFF
--- a/src/essence/Basics/Layers_/Layers_.js
+++ b/src/essence/Basics/Layers_/Layers_.js
@@ -2130,5 +2130,3 @@ function parseConfig(configData, urlOnLayers) {
 }
 
 export default L_
-
-window.L_ = L_
--- a/src/essence/Basics/Layers_/Layers_.js
+++ b/src/essence/Basics/Layers_/Layers_.js
@@ -1948,20 +1948,14 @@ var L_ = {
     //Takes in a config layer object
     globeLithoLayerHelper: async function(s) {
         if (L_.Globe_) {
-            // Hold the previous layer state
-            const prevState = L_.toggledArray[s.name]
-
             // Only toggle the layer to reset if the layer is toggled on,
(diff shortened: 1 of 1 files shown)


src/essence/Basics/Layers_/Layers_.js
```diff
@@ -1937,6 +1945,26 @@ var L_ = {
...
+            // because if the layer is toggled off, it is not on the globe
+            if (prevState) {
+                // Temporarily set layer to on
+                L_.toggledArray[s.name] = false
+
+                // Toggle the layer so its drawn in the globe
+                await L_.toggleLayer(s)
+
+                // Reset the previous layer state
+                L_.toggledArray[s.name] === prevState
```
tariqksoliman, line 1964: Can we try to get rid of this line? `L_.toggleLayer()` should already be doing this on its own terms.  

Also, if `L_.toggleLayer()` doesn't need the await, I'd get rid of it.

tariqksoliman (changes requested):

src/essence/Basics/Layers_/Layers_.js
ac-61: Fixed in 5f38e6800938884468eefc5b3bb1c2f1452dee52!

src/essence/Basics/Layers_/Layers_.js
```diff
@@ -2108,3 +2130,5 @@ function parseConfig(configData, urlOnLayers) {
 }
 
 export default L_
+
+window.L_ = L_
```
tariqksoliman, line 2134: Great! But now this.

tariqksoliman (changes requested):

tariqksoliman (approved):
LGTM!

Issue #185, opened 2022-05-19
Annotations in 3D


tariqksoliman:
Will include some visibility cutoff improvements as well.

PR #188, opened 2022-05-24
#185 Annotations in 3D

tariqksoliman:
Updates and integrates LithoSphere v1.5.0. Adds support for per-feature visibility cutoffs with `minZoom` and `maxZoom` style properties. The previous `visibilityCutoff` field is deprecated but still supported.


tariqksoliman:
Closes #185

PR #189, opened 2022-05-26
Description should be updated to match example code

davetsay:
While copy pasting to set a label property, I chose poorly. The example code is correct and the description is not.


tariqksoliman (approved):
Ooo, nice catch!
Normally we should always merge into development first, but since this is such a minor change, master is fine..

PR #205, opened 2022-07-18
Set initial layer times

### LINKED ISSUE #199: Initial Layer Times
- Allow initial start and end times to be settable via a deep link in the url. No empty or default time based queries should be performed if these are set.
- Additionally fix how TIME, STARTTIME and ENDTIME are injected into WMS layers. These parameter names should come from the configured url (i.e. `whatever-name_I-want={starttime}`) instead of being hardcoded. I don't remember but those hardcoded parameters might not also properly apply at first either. See https://github.com/NASA-AMMOS/MMGIS/blob/development/src/essence/Ancillary/TimeControl.js#L208

ac-61:
This closes #199.

### DIFF
--- a/documentation/pages/markdowns/Deep_Linking.md
+++ b/documentation/pages/markdowns/Deep_Linking.md
@@ -124,6 +124,22 @@ If no `mission` is set in the URL, there exists only one configured mission and 
 - _dependencies_: mission is unset
 - _unset:_ The landing page will redirect to the only mission if there exists only one configured mission
 
+### startTime=<br>`<start time>`
+
+Sets the start time for the time control feature.
+
+- _form:_ `<string>`
+- _dependencies_: Time control
+- _unset:_  Will default to current time if time control feature is enabled
+
+### endTime=<br>`<end time>`
+
+Sets the end time for the time control feature.
+
+- _form:_ `<string>`
+- _dependencies_: Time control
+- _unset:_  Will default to current time if time control feature is enabled
+
 ## Examples
 
 #### General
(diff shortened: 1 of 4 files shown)


davetsay:
I have a couple of questions.

Are the time params setting the global mmgis time context? Will there be anyway to set a layer time context? Our need for setting individual layers' times would be for things like plans, which would be fixed times in the future most likely. This seems like an edge case we can work around, but curious if this was given consideration.

Will we ever be getting time params back from `writeCoordinateUrl` and/or the copy link feature? I don't think it matters much to us either way, but we might need to account for replacing params instead of appending if they exist.

ac-61:
@davetsay 
> Are the time params setting the global mmgis time context? Will there be anyway to set a layer time context? Our need for setting individual layers' times would be for things like plans, which would be fixed times in the future most likely. This seems like an edge case we can work around, but curious if this was given consideration.

The deep link from the url currently sets the global MMGIS time. We could discuss further if individual layer times can be set via deep links. 

> Will we ever be getting time params back from writeCoordinateUrl and/or the copy link feature? I don't think it matters much to us either way, but we might need to account for replacing params instead of appending if they exist.

Not at the moment.

Pinging @tariqksoliman for opinions on the above :grin:

tariqksoliman:
1. If deep linked times per layer are an edge case and have a work around, I'm tempted to ignore them now unless it's specifically requested.

2. I think we should (and that it's within the realm of this ticket) write the current global time (if any and if enabled) back into the url when users "copy link".

ac-61:
> I think we should (and that it's within the realm of this ticket) write the current global time (if any and if enabled) back into the url when users "copy link".

@davetsay The latest update to this PR should fulfill the above requirement.

tariqksoliman (approved):
LGTM!

PR #214, opened 2022-08-09
Add new event listeners for MMGIS API

### LINKED ISSUE #203: API listener needed for switching tools (tool status)
Use Case
On click, we pivot location data to instrument data. We do not want to pivot when using the drawing tool, but we do want to pivot when using the layer tool

ac-61:
This closes #203. This adds a listener for changing tools and adds a listener for layer visibility changes. The two new options are: `toolChange` and `layerVisibilityChange`. Both options have functions that are called with an `event` parameter. This also adds a `getActiveTool` function to get the active tool.

### Example
```
function mmgisListener(event) {
    console.log('event', event)
}
window.mmgisAPI.addEventListener('toolChange', mmgisListener)
```

### DIFF
--- a/documentation/pages/markdowns/JavaScript_API.md
+++ b/documentation/pages/markdowns/JavaScript_API.md
@@ -380,7 +380,7 @@ window.mmgisAPI.updateLayersTime()
 
 ### addEventListener(eventName, functionReference)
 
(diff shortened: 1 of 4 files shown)


tariqksoliman (approved):
Tried it out and it LGTM!

PR #215, opened 2022-08-09
Allow mmgisAPI to individually reload any layer (not just time layers)

### LINKED ISSUE #202: API method `reloadLayer()` should work for non time based layers
Use Case
Raster layers for instrument data should be reloadable, so the user can see the most current data. The layers shouldn't be time based, as the most current raster should always be shown.

ac-61:
This closes #202. This makes it so any layer can be refreshed with the `reloadLayer(layer)` function.

### DIFF
--- a/src/essence/Ancillary/TimeControl.js
+++ b/src/essence/Ancillary/TimeControl.js
@@ -202,36 +202,34 @@ var TimeControl = {
         if (typeof layer == 'string') {
             layer = L_.layersNamed[layer]
         }
-        if (layer.time && layer.time.enabled == true) {
-            var layerTimeFormat = d3.utcFormat(layer.time.format)
-            layer.time.current = TimeControl.currentTime // keeps track of when layer was refreshed
+        var layerTimeFormat = d3.utcFormat(layer.time.format)
+        layer.time.current = TimeControl.currentTime // keeps track of when layer was refreshed
 
-            if (layer.type == 'tile') {
-                TimeControl.setLayerWmsParams(layer)
+        if (layer.type == 'tile') {
+            TimeControl.setLayerWmsParams(layer)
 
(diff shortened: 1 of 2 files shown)


src/essence/Ancillary/TimeControl.js
```diff
@@ -202,36 +202,34 @@ var TimeControl = {
...
-                        '{endtime}',
-                        layerTimeFormat(Date.parse(layer.time.end))
-                    )
-                // refresh map
-                if (L_.toggledArray[layer.name] || evenIfOff) {
-                    await Map_.refreshLayer(layer)
-                }
-                // put start/endtime keywords back
-                layer.url = originalUrl
+        var layerTimeFormat = d3.utcFormat(layer.time.format)
```
tariqksoliman, line 205: Does this break if a layer does not have `layer.time`? It might make more sense to move `if (layer.time && layer.time.enabled == true)` around new L209 and around L217 -> L226. Thoughts?

tariqksoliman (changes requested):

src/essence/Ancillary/TimeControl.js
ac-61: Thanks! I fixed this in b1b8a1f8b16e2f1bba984464290a7319649d9e51. Originally, I thought it wouldn't matter since time is generally added below but better safe than sorry:
https://github.com/NASA-AMMOS/MMGIS/blob/0a5a6348c276b560a33a5a36b110cde8bcb8c0e3/config/js/config.js#L2025-L2038.

tariqksoliman: Oh right! Though do remember since we have quite a few MMGIS instances and older configurations around, that we can't assume things are set.

tariqksoliman (approved):
LGTM!

PR #227, opened 2022-09-07
Make sure WMS parameters are set correctly

ac-61:
This makes sure the layer WMS parameters are set without calling reloadLayer.


tariqksoliman (approved):
LGTM

PR #229, opened 2022-09-08
Set WMS parameters when calling setTime function

ac-61:
Similar fix to #227. This makes sure the WMS parameters are set correctly when the `setTime` function is called without calling `reloadLayer`.


tariqksoliman (approved):
LGTM!

PR #230, opened 2022-09-09
Update geojson data when removing features from vector layers

ac-61:
This fixes a bug where image icons were not removed from the map for vector layers when calling API functions such as `keepLastN`.


tariqksoliman (approved):
LGTM and thank you!

PR #233, opened 2022-09-13
Fix onClick event listener bug

ac-61:
This fixes the following bug:
> I’m setting up “onClick” and “toolChange” event listeners after MMGIS loads. It looks like when I switch to the draw pane, the onClick event listener stops working (which isn’t a problem for us), but when I switch back to any of the other panes (layers, legend, etc.) the onClick event listener doesn’t resume firing on clicks. I can send a screen recording if this doesn’t make sense.

This fixes the bug because `d.start` is an undefined function so it was removing _all_ of the click event listeners on the Leaflet map.

### DIFF
--- a/src/essence/Tools/Draw/DrawTool_Drawing.js
+++ b/src/essence/Tools/Draw/DrawTool_Drawing.js
@@ -586,7 +586,6 @@ var drawing = {
 
             d.stopclick = false
 
-            Map_.map.off('click', d.start)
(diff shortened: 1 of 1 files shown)


tariqksoliman (approved):
Nice catch! I would think `off('click', null)` would do nothing and only `off('click')`/`off('click', undefined)` would clear the event entirely.

PR #240, opened 2022-09-27
Fix TimeControl so WMS information is set before layers are loaded

ac-61:
This should fix an issue where the WMS layer times are not set before the `mmgisAPI.onLoaded` callback function is called.

### DIFF
--- a/src/essence/Ancillary/TimeControl.js
+++ b/src/essence/Ancillary/TimeControl.js
@@ -90,6 +90,8 @@ var TimeControl = {
         if (L_.configData.time.visible == false) {
             TimeControl.toggleTimeUI(false)
         }
+    },
+    fina: function () {
         initLayerTimes()
     },
     toggleTimeUI: function (isOn) {
--- a/src/essence/essence.js
+++ b/src/essence/essence.js
@@ -178,14 +178,14 @@ var essence = {
         //Make the map
         if (swapping) Map_.clear()
 
+        //Make the time control
+        TimeControl.init()
+
         Map_.init(this.fina)
 
         //Now that the map is made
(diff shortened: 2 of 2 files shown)


tariqksoliman:
Can you explain the change more and why things are ordered the way they are now? Why does splitting and moving the TimeControl UI initialization before the Maps_'s initialization help anything? Map_ calls `TimeControl.updateLayersTime` (which in turns calls `TimeControl.setLayerWmsParams`) but those function don't seem to depend on anything done in `TimeControl.init`

ac-61:
@tariqksoliman I don't think matters as much when we initialize TimeControl's UI (so we could move `TimeControl.init()` back down. The current development branch calls `TimeControl.updateLayersTime` before `TimeControl.initLayerTimes` (`TimeControl.initLayerTimes` is in `TimeControl.init`, which is called _after_  `Map_.init()`).

The code in the development branch is in the following order:
- essence
    - Map_.init
        - Map_.allLayersLoaded()
            - essence fina
                - mmgisAPI_.fina(Map_) (call back to `mmgisAPI.onLoad()`)
        - TimeControl.updateLayersTime()
    - TimeControl.init()
        - TimeControl.initLayerTimes()

But what I think we want is:
- essence
    - TimeControl.init()
    - Map_.init()
        - Map_.allLayersLoaded()
            - essence fina
                - TimeControl.initLayerTimes()
                - mmgisAPI_.fina(Map_) (call back to `mmgisAPI.onLoad()`)
        - TimeControl.updateLayersTime()

tariqksoliman (approved):
Great, thanks and looks good!

PR #246, opened 2022-09-29
Add WebSockets feature to enable pushing updates to clients

ac-61:
This adds a feature to open a WebSocket on the MMGIS server to broadcast to clients upon updates with the REST API. Currently, this will only broadcast to clients with calls to the `addLayer`, `updateLayer`, and `removeLayer` and when saving the MMGIS configuration page. This feature can be enabled by setting the environment variable `ENABLE_MMGIS_WEBSOCKETS` to true.

When users make calls using the rest API, a notification of updates to the clients are displayed in a button the right side of MMGIS and a modal appears with the updates when the button is clicked. The `addLayer`, `updateLayer`, and `removeLayer` REST API calls have a `forceClientUpdate` parameter that will update the connected clients without a button/modal.

When saving the MMGIS configuration with new updates, a button on the client side will appear to tell the users they need to refresh the page. A modal will appear when the button is pressed to confirm the user would like to refresh the page.


tariqksoliman (changes requested):
Looks great! Very exciting to see `updateLayer` placing the layers in the correct position.

I did just make commit to this branch to fix an issue where the page would crash if you didn't have ENABLE_MMGIS_WEBSOCKET in your .env.


Some changes and fixes:

### Logging:
- Remove "Started websocket connection..." client log
- "Fix" `Error parsing data from MMGIS websocket: SyntaxError: Unexpected token 'H', "Hello, I a"... is not valid JSON` log
- Remove Backend "received in websocket API: ..." log as it's very long. (Keep info logs tiny and error logs can be full)
	
### UI:
- When "Map Zoom Control" is unchecked in the /configure Look tab the Reload button is misplaced
- Reload Page should get an Are you sure modal. (I did very recently add https://github.com/NASA-AMMOS/MMGIS/blob/development/src/essence/Ancillary/ConfirmationModal.js if you want to use that (It's just `ConfirmationModal.prompt("message", (isYes) => {} )`))
- The tooltips are a bit misplaced. Since we're using modals now I think these tooltips are far less important. (Maybe just the html "title" is enough alongside the modal.)
- In the Update Map Layers Modal, rename "Type" -> "Action" and make the addLayer, updateLayer, removeLayer pretty ("Add Layer", ...)
	
### Websocket:
- A "failed" Configure API call to /addLayer still notifies the client of the (false) change
- A Configure API call to /addLayer for a separate mission still shows something changed (Client should check whether the config change is from the current mission)

tariqksoliman (approved):
Nice and LGTM!

PR #252, opened 2022-10-05
Fix deep link times for vector layers

ac-61:
We need to make sure to set the times before the non WMS layers are created.


tariqksoliman (approved):
Ah, so the issue was with the multiple objects to define the layers. They (i.e. L_.layersNamed) should be read-only and any writes should be in the main parseConfig but this should be great for now.

Thanks!

PR #244, opened 2022-09-29
Add toggleLayer function to MMGIS API

ac-61:
This adds a function to the MMGIS API to toggle a layer's visibility state.

Examples:
```javascript
# Set 'Layer 1' to be visible
mmgisAPI.toggleLayer('Layer 1', true)

# Toggle the visibility state of 'Layer 1'
mmgisAPI.toggleLayer('Layer 1')
```

PR #260, opened 2022-10-17
Fix toggleLayer bug in MMGIS API

ac-61:
This fixes a bug introduced in #244, where the second parameter in `mmgisAPI.toggleLayer('Layer 1', false)` would toggle a layer on, despite the value being `false`.


jtroberts (approved):

ac-61:
@tariqksoliman If you get a chance, could you please review this PR? Thanks!

tariqksoliman (approved):
LGTM!

PR #281, opened 2022-11-30
Fix environment variable in sample.env for enabling websockets

ac-61:
This fixes a typo in the sample.env file for enabling the websockets feature.


tariqksoliman (changes requested):
Do change it in the docs too: https://github.com/NASA-AMMOS/MMGIS/blob/development/docs/pages/Setup/ENVs/ENVs.md#enable_mmgis_sockets

tariqksoliman (approved):
Thanks!

PR #288, opened 2022-12-09
Update README to note the minimum required Node.js version

ac-61:
The `sharp` package requires node >= 14.15.0.


tariqksoliman:
Outdated. Requirements are now Node.js v16.13.2+ (including versions v18+)

PR #295, opened 2022-12-13
Add feature to reconnect WebSocket from client to server

ac-61:
This feature allows the WebSocket on the client side to reconnect to the server if it is ever disconnected. The client checks if the WebSocket connection is open every minute and if the connection is closed, it will attempt to retry to connect. If the reconnection fails, the interval between reconnection attempts doubles for every failure.

A toast is displayed both when connected and disconnected from server WebSocket.


tariqksoliman (changes requested):
Main functionality looks perfect!

Could you get rid of the first connection toast since it's kind of annoying. Only show Successfully Connected if it had previously failed. Also shift the toasts to be only ~5 pixels from the right side.

ac-61:
@tariqksoliman I made the requested changes!

src/essence/essence.js
```diff
@@ -125,6 +126,147 @@ $(document.body).keydown(function (e) {
...
+            console.log('Websocket connection opened...')
+
+            UserInterface_.removeLayerUpdateButton()
+
+            M.Toast.dismissAll()
+
+            if (essence.webSocketRetryInterval > essence.initialWebSocketRetryInterval) {
+                M.toast({
+                    html: "Successfully connected to WebSocket",
+                    displayLength: 160000000,
```
tariqksoliman, line 174: Nice! And this is a lot of milliseconds.

ac-61: Crud, I would say something about Monday but it's Tuesday now! Fixed with 3b4cdd011e0296db05cdc2d80ba92e358eeb2164!

tariqksoliman (approved):
LGTM!

PR #310, opened 2023-01-20
Fix function call order bug in addNewLayer function

ac-61:
This fixes a bug in the `addNewLayer` function, which would crash if adding a new layer via the Rest API and then updating the client either manually or with forced updates. We should look for the new layer's UUID after parsing the database information with`parseConfig`.


tariqksoliman (approved):
Nice catch!

PR #320, opened 2023-01-30
Add function to overwrite elements displayed in LegendTool

ac-61:
This adds a function called `overwriteLegends` in the LegendTool to overwrite the contents displayed in the LegendTool. This can be useful when used with the "toolChange" event listener in the MMGIS API.

The `overwriteLegends` function takes in an array of objects, where each object must contain the following keys:
legend, layerUUID, display_name, opacity. The value for the legend key should be in the same format as what is stored in the layers data under the `_legend` key (i.e. `L_.layers.data[layerName]._legend`). The value for the layerUUID and display_name keys should be strings and opacity should be a number between 0 and 1.

### DIFF
--- a/src/essence/Tools/Legend/LegendTool.js
+++ b/src/essence/Tools/Legend/LegendTool.js
@@ -19,6 +19,7 @@ var LegendTool = {
     destroy: function () {
         this.MMWebGISInterface.separateFromMMWebGIS()
     },
+    overwriteLegends: overwriteLegends,
 }
 
 //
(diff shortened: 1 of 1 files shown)


tariqksoliman:
Cool! But just wondering: should `overwriteLegends` be exposed in the `mmgisAPI` as well?

ac-61:
@tariqksoliman Great idea! I added `overwriteLegends` in the `mmgisAPI`.

tariqksoliman (approved):
LGTM!

Issue #313, opened 2023-01-25
Expose as much as possible to the mmgisAPI


tariqksoliman:
Such as all the utility formulas and everything (reasonable) in Layers. Document as well.

The-Outrider:
validateTokenKey would also be nice.

tariqksoliman:
@The-Outrider Could you explain that more or provide a use case? Are the tokens here the API Tokens used to interact with the Configure API.

(
For some definitions:
- mmgisAPI = JavaScript API = applications wrapping MMGIS would use this to interface with the client-instance of MMGIS = called with JavaScript
- Configure API = backend REST API that requires auth = modifies the MMGIS configuration for the whole mission for all users = called with curl/fetch/whatnot  

)

The-Outrider:
Hey Tariq,
![Screen Shot 2023-02-09 at 09 45 51](https://user-images.githubusercontent.com/6240904/217863197-9e9c721a-aa7f-4a7b-81fe-3eceea5468af.png)

Yah its for the Configure API, the basic idea is to have a check if the API token is set and is still valid. This would displayed on the frontend.  (See screenshot)

tariqksoliman:
Okay, I see what you mean now.

tariqksoliman:
@The-Outrider Hi! See what I've added here: https://github.com/NASA-AMMOS/MMGIS/pull/330 and of course I can always add more on request.

PR #330, opened 2023-02-10
#313 Expose as much as possible to the mmgisAPI

tariqksoliman:
- Adds all utility functions to the mmgisAPI along with docs: https://nasa-ammos.github.io/MMGIS/apis/javascript/utils
- Adds `addLayer` to add MMGIS controlled layers to the mmgisAPI: https://nasa-ammos.github.io/MMGIS/apis/javascript/#addlayerlayerobj-placement
- Adds `removeLayer` to remove MMGIS controlled layers to the mmgisAPI: https://nasa-ammos.github.io/MMGIS/apis/javascript/#removelayerlayeruuid
- Adds the `logged_in` endpoint:  https://nasa-ammos.github.io/MMGIS/apis/backend#get-apiuserslogged_in
- Fixes a few bugs in Formulae_ found while writing its documentation
- Moves around a few functions to where they make better sense


The-Outrider:
the logged_in was the function you mentioned regarding #313 is for the frontend api isn't it? Not the API key?

tariqksoliman:
It's a GET request that checks whether the current user is logged in to the client.

It's not the long-term Admin API token that's made on the /configure for programmatically interacting with the configuration endpoints. Any request to those would say something like "not authenticated" anyway you you need to check that as well.

Issue #350, opened 2023-03-07
[New Feature]: Optional Info button for projects that desire a link to general project info 


jtroberts:
### Checked for duplicates

Yes - I've already checked

### Alternatives considered

Yes - and alternatives don't suffice

### Describe the feature request

I need an optional Info button for projects that desire a link to general project info. This is in addition to the existing Help button which we're using for tool instructions.

PR #351, opened 2023-03-07
Optional Info button for projects that need a link to project info #350

jtroberts:
## Purpose
Optional Info button for projects that desire a link to general project info #350
## Proposed Changes
- [ADD] Info button
## Issues
- #350 
## Testing
- Tested adding Info URL to config page and button appears on sidebar with correct link
- Tested removing Info URL and button is hidden from sidebar

### DIFF
--- a/config/js/config.js
+++ b/config/js/config.js
@@ -625,7 +625,7 @@ function initialize() {
                 }
                 if (
                   cData.look &&
-                  (cData.look.info == true || cData.look.info == null)
+                  (cData.look.info == true)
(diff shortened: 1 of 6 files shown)


tariqksoliman (approved):
Nice and LGTM!

Some minor and optional thoughts if you want:
- Hide the i and ? icons if not url is configured
- Default the info to be off in the configuration (when it's unset)
- Use a solid circle information icon (because the info tool already uses the same i icon)

PR #375, opened 2023-05-08
chore(config): Inline import from same namespace

evenstensberg:
## Purpose
- A lil bit syntax sugar


tariqksoliman (approved):
LGTM and thanks again

PR #384, opened 2023-05-26
fix(imports): destructure modules

evenstensberg:
## Purpose
Makes source cleaner

## Proposed Changes
N/A
## Issues
N/A
## Testing

Not tested on my machine, needs to be tested locally. In theory should work


tariqksoliman (approved):
LGTM and works, thanks!

PR #385, opened 2023-05-26
chore(naming): use standardized name

evenstensberg:
## Purpose
Normal naming for running your app through node/bash/sh is `scripts/<name>.ext`
## Proposed Changes
N/A
## Issues
N/A
## Testing
N/A


tariqksoliman (approved):
Thanks again. (I believe the `run/` folder was named this way because, before moving to node, there was already a `scripts/` folder. "run" also feels like a better name overall too.)

PR #388, opened 2023-06-23
Add event listener for toggleSeparatedTool in MMGIS API

ac-61:
## Purpose
- This adds a listener so we can listen to the toggle state of separated tools (introduced in #379).

## Issues
- Enhancement to #379

## Testing
This was tested by pasting the following into the dev console and then toggling the Legend Tool on and off.
```
function mmgisListener(event) {
  console.log("event", event);
}

window.mmgisAPI.addEventListener("toggleSeparatedTool", mmgisListener);
```


tariqksoliman (approved):
Cool and thank you!

PR #393, opened 2023-07-12
Added an option to the LegendTool to automatically display on start

jtroberts:
## Purpose
- Some projects like having the legend visible and expanded when the map loads.
## Proposed Changes
- This PR adds an option to the LegendTool to automatically display the legend on start.
## Testing
- Tested with LegendTool turned on/off.
- Tested with displayOnStart set to true/false/null/etc.
- Legend behaves as expected.


tariqksoliman (approved):
Nice and thank you!

PR #394, opened 2023-07-13
Added max-width to legend to prevent long text from filling screen

jtroberts:
## Purpose
- I have legend values with long descriptive text that are expanding the legend to fill nearly the entire screen.
## Proposed Changes
- Added a max-width property to limit the expansion of text.
## Testing
- Tested different legends with various css values to find a good compromise.


tariqksoliman (approved):
Looks good. It might be worthwhile to use `text-overflow: ellipsis;` and a `title` attribute too.

jtroberts:
The ellipsis with title looks a lot better than the scrollbar. Thanks!

PR #409, opened 2023-08-07
Allow the legend tool to be right justified on the screen

jtroberts:
## Purpose
- Mission has requested that the floating legend be moved to the right side of the screen to maximum screen space.
## Proposed Changes
- Added custom option to allow legend tool to be right justified on the screen
## Testing
- Added configuration option to legend tool and legend is moved to the right
- Removed configuration option to make sure tool still works as before without it

### DIFF
--- a/docs/pages/Tools/Legend/Legend.md
+++ b/docs/pages/Tools/Legend/Legend.md
@@ -9,11 +9,15 @@ parent: Tools
 
 A layer can be configured with a legend by pointing its Legend field to a .csv file or by including a JSON `legend` array into the layer's Raw Variables. The Legend Tool renders symbologies and gradient scales for any properly configured on layer.
 
-On the Configure page, under Tools, you can specify if the expanded legend should automatically be displayed on start:
+On the Configure page, under Tools, you can specify additional options:
+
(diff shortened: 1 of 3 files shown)


tariqksoliman (approved):
Nice! 
A couple of things:
- Include an example Raw Variables in "descriptionFull" in `src/essence/Tools/Legend/config.json`
   - See other tool config.jsons for examples. This makes it so that in the Tools tab of the /configure page, you can click on the tool's info icon and see it as an example
- Not as critical either but the right hand side of the screen is a bit tricky: The map Zoom In and Out are option (see /configure look tab). And when Zoom In and Out are enabled, the right side Legend covers the websockets refresh/reload page icon. It's a bit trickier still because the existing right-side icons are often fixed to the panel (map, viewer, globe) and not the screen. Again it's minor, so feel free to merge. Likely all the right-side icons need to be drawn more intelligently.

jtroberts:
Added the example in descriptionFull. I'm going to leave the right hand side icon updates for another day. I'm not familiar with how to enable the websockets icon and it does get tricky with the globe and viewer on.

PR #434, opened 2023-10-05
Spaceapps

andresinho20049:
## Purpose
- run project with docker without difficulties

```
db:
    image: postgis/postgis:10-2.5-alpine
    env_file: .env
    ports:
      - 5433:5432
    restart: on-failure
    volumes:
      - mmgis-db:/var/lib/postgresql/data
      - ./Database:/usr/src/app/Database
      - ./Database/PgSQL_queries/mmgisDB_queries.sql:/docker-entrypoint-initdb.d/mmgisDB_queries.sql
```

The base of this image is the postgres image which has a docker entrypoint to start along with the container.

Bash and SQL scripts;

## Proposed Changes
- [ADD] Docker entrypoint with init DB for started config
- [CHANGE] Run mode for developer
## Testing
- Test Container to integrate and test unit


andresinho20049:
# Pull request 

I'm participating in the space apps challenge and I've been studying a lot about the EMIT challenge.

/change

tariqksoliman:
Hi @andresinho20049 and thanks for the PR!

A couple of things:
- Please PR into the `development` branch. `master` is solely for the latest release.
  - (I agree there should be a contributing guide that says this.)
- Everything in the /Database folder is legacy, unused and probably should be deleted.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:16
+FROM node:16.13.2@sha256:7c49a64aba86dd483aa874bc0230c07f282e20741a7c66e426970ecafc149a38 as base
```
tariqksoliman, line 1: We should probably upgrade this to Node v20.7

scripts/server.js
```diff
@@ -717,11 +717,11 @@ setups.getBackendSetups(function (setups) {
...
   // We're only doing this for dev because we're assuming
-  // build will also call this.
-  if (process.env.NODE_ENV === "development") {
-    console.log(chalk.cyan("Updating Tools...\n"));
-    updateTools();
-  }
+  // // build will also call this.
+  // if (process.env.NODE_ENV === "development") {
+  //   console.log(chalk.cyan("Updating Tools...\n"));
+  //   updateTools();
```
tariqksoliman, line 723: This is in fact needed. MMGIS can link into sub-repos of private tools and this `updateTools` functions scans for them and plugs them in.

scripts/server.js
```diff
@@ -730,12 +730,12 @@ setups.getBackendSetups(function (setups) {
...
   // Start listening for requests.
   httpServer.listen(port, (err) => {
-    if (process.env.NODE_ENV === "development") {
+    if (process.env.NODE_ENV === "development") 
       setTimeout(setupDevServer, 2000);
-      app.get("/", ensureUser(), (req, res) => {
-        res.redirect(`http://localhost:${port + 1}`);
-      });
-    } else {
+    //   app.get("/", ensureUser(), (req, res) => {
```
tariqksoliman, line 735: Removing this DEV vs PROD distinction is breaking hot-reloading in DEV (because PROD is solely relying on the /build folder and scripts)

Dockerfile
```diff
@@ -18,11 +18,21 @@ COPY package*.json ./
...
+FROM base as prod
+
 # Bundle app source
 COPY . .
 
 # build
 RUN npm run build
 
 EXPOSE 8888
-CMD [ "npm", "run", "start:prod" ]
```
tariqksoliman, line 28: What's starting the application if this is removed?

docker-compose.yml
```diff
@@ -1,22 +1,31 @@
...
       - ./Missions:/usr/src/app/Missions
+      - ./scripts:/usr/src/app/scripts
+      - ./src:/usr/src/app/src
+
   db:
     image: postgis/postgis:10-2.5-alpine
     env_file: .env
     ports:
-      - 5432:5432
+      - 5433:5432
```
tariqksoliman, line 24: What's the reasoning behind port 5433?

docker-compose.yml
```diff
@@ -1,22 +1,31 @@
...
     restart: on-failure
     volumes:
       - mmgis-db:/var/lib/postgresql/data
+      - ./Database:/usr/src/app/Database
```
tariqksoliman, line 28: Again, the MMGIS ./Database directory is obsolete.

docker-compose.yml
```diff
@@ -1,22 +1,31 @@
...
+      - ./Database/PgSQL_queries/mmgisDB_queries.sql:/docker-entrypoint-initdb.d/mmgisDB_queries.sql
```
tariqksoliman, line 29: Again, the MMGIS ./Database directory is obsolete.

Dockerfile
andresinho20049: OK, noted. I set the version that was in the NVMRC file, I will update.

scripts/server.js
andresinho20049: I offer my sincerest apologies for the error that occurred in my recent upload. My intention was to upload changes to my fork specifically, but by mistake, I uploaded them to the main repository instead. These changes were designed to simplify the process of running the project in development mode on my personal device, as I am not currently part of the NASA team. Regrettably, I also submitted a pull request to the master branch instead of creating a new branch named "spaceapps."

Upon reflection, I now understand that I should have exercised greater caution with respect to the timing of my pull request. I had previously created a new branch and made the necessary alterations in separate commits. However, during the pull request process, I became confused and made an error. I am deeply sorry for any inconvenience or confusion my actions may have caused.

I assure you that this was an unfortunate mistake, and I am committed to ensuring that it will not occur in the future. Once again, I apologize for any inconvenience caused.

scripts/server.js
andresinho20049: This change was not supposed to be in the pull, I modified it to run locally

Dockerfile
andresinho20049: I left the entrypoint call in compose, because in this case I left the environment thought of as a development environment

docker-compose.yml
andresinho20049: I'm sorry, I already had another Postgres database running and that's why I changed it to avoid local conflicts.

docker-compose.yml
andresinho20049: Ok! But it was good because I was able to run EMIT locally and load the prod configurations.

scripts/server.js
tariqksoliman: No worries whatsoever! This is what PRs are for too.

scripts/server.js
tariqksoliman: Could you also detail the steps you've been taking to run this locally? We are interested in simplifying the initial installation and supporting a variety setups. Often we develop simply with `npm start` and only use Docker for final production images if the customer wishes to go that route.

andresinho20049: Within the dockerfile, each time the FROM statement is included, a new layer of the image is created, so I divide the construction of the docker image with multi-stage.

### Normally
   - the bases
   - as builder
   - as develop
   - and final version

The big trick when running in development is to configure the volume, basically building a volume pointing to the src and in the script instead of starting with node, using nodemon for example, in java using Dev tools dependency, then it is possible to run in development.

And in the final version, I only copy the build / out file and only the run time in this image, unlike the dev one, here the intention is to leave just one image with instructions to run the package and as light as possible.

---
My sample of Dockerfile

```
########### Stage BUILD ###########
FROM maven:3.6.3-jdk-8-slim as builder

# Define Workdir
WORKDIR /opt/app
#Copy pom.xml to container
COPY pom.xml ./

RUN mvn dependency:go-offline

COPY ./src ./src
RUN mvn clean
RUN mvn install package

########### Stage RUN ###########
FROM eclipse-temurin:8-jre-alpine@sha256:f6485e494f9927cb6a51e23134b3556c62c32c3df20fc4317f7f493343726977 as runner
SHELL ["/bin/sh", "-c"]

RUN mkdir -p /etc/skel/

RUN echo 'history -c \
/bin/rm -f /opt/remote/.mysql_history \
/bin/rm -f /opt/remote/.history \
/bin/rm -f /opt/remote/.bash_history' >> /etc/skel/.logout

RUN echo 'set autologout = 30 \
set prompt = "$ " \
set history = 0 \
set ignoreeof' >> /etc/skel/.cshrc

RUN cp /etc/skel/.cshrc /etc/skel/.profile

# dumb-init occupies PID 1 and takes care of all the responsibilities.
RUN apk add dumb-init

# Define Workdir
WORKDIR /opt/app

# Add user javauser based minumum privileges
RUN addgroup -S -g 1000 javagroup && \
    adduser -S -s /bin/false -G javagroup -u 999 javauser

RUN echo "javauser:secretPass" | chpasswd

# Copy target from build stage
COPY --from=builder /opt/app/target/AuthService.war ./

RUN chmod 0 /home \
    && chmod 0 /var \
    && chmod 0 /usr \
    && chmod 0 /dev \
    && chmod 0 /mnt \
    && chmod go-rw /etc \
    && chmod 0 /etc/ssl \
    && chmod 0 /etc/skel \
    && chmod a+rwx /usr/bin/dumb-init

RUN chown -R javauser:javagroup /opt/app
USER javauser

ENTRYPOINT ["dumb-init", "java", "-jar", "AuthService.war"] # I will create an entrypoint

```

> In the case of this image presented above, I run it through compose pointing the target to the builder version and I configure a volume in the ./src folder

tariqksoliman:
Hi again and I'm closing this issue because we've since fleshed out our docker-compose.sample.yml and Dockerfile. Any recommendations to them would be much appreciated too. Thanks!

Issue #452, opened 2023-11-13
[New Feature]: Floating separated Identifier Tool


jtroberts:
### Checked for duplicates

Yes - I've already checked

### Alternatives considered

Yes - and alternatives don't suffice

### Related problems

I'm frustrated when I have tons of layers and I use the Identifier tool to inspect values, but I need to constantly flip back and forth between the Layer tool and Identifier.

### Describe the feature request

I want to make the Identifier tool a separated floating tool just like the Legend tool so that I can keep the Layer pane open when inspecting values.

PR #453, opened 2023-11-13
#452 Updated Identifier tool to be a separated floating tool

jtroberts:
## Purpose
- Updated Identifier tool to be a separated floating tool
## Issues
- #452 
## Testing
- Currently being used by one mission and works as expected with tools on the right-side of the UI
- Clicking on tool icon activates tool, allows user to hover over data to see values, tool remains activated with Layer tool open and user switching between layers, clicking again on tool icon deactivates tool

- Need to test different UI combinations with other separated tools
- We may possibly want an option to make the tool not be separated to maintain compatibility

### DIFF
--- a/src/essence/Tools/Identifier/IdentifierTool.js
+++ b/src/essence/Tools/Identifier/IdentifierTool.js
@@ -467,6 +467,7 @@ function interfaceWithMMWebGIS() {
     //tools.html( markup );
 
     //Add event functions and whatnot
+    var previousCursor = d3.select('#map').style('cursor')
     d3.select('#map').style('cursor', 'crosshair')
 
     Map_.map.on('mousemove', IdentifierTool.idPixelMap)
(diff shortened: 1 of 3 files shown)


tariqksoliman (changes requested):
Looks good to me but it really needs to be backwards compatible too. Before I saved the configuration again (to make the IdentifierTool default to separated), clicking on the IdentifierTool in the toolbar made it impossible to open any other tool afterwards.

tariqksoliman (approved):
Looks good to me!
Though I keep seem to justify it to the right unless the legend is also justified to the right.

jtroberts:
I added support to allow the Identifier and Legend tools to be on different sides if specified. We may need a more scalable though, if we enable more separated tools in the future.

tariqksoliman (approved):
Okay, looks completely good to me now then.

Issue #546, opened 2024-05-30
[New Feature]: Implement a way to download the source data for raster layers


jtroberts:
### Checked for duplicates

Yes - I've already checked

### Alternatives considered

Yes - and alternatives don't suffice

### Describe the feature request

Mission has a need to allow users to download the source data file (e.g. GeoTIFF) for raster layers. Specifically, a download button in the layer tool similar to vector layers.

PR #547, opened 2024-05-30
Added downloadURL variable to allow users to download raster layers #546

jtroberts:
## Purpose
- Mission has a need to allow users to download the source data file (e.g. GeoTIFF) for raster layers
## Proposed Changes
- Added downloadURL variable to allow users to download raster layers from the Layer Tool
## Issues
- #546 Implement a way to download the source data for raster layers 
## Testing
- Configured several layers with the downloadURL variable, download button shows up next to the layer in the tool, source file is able to be downloaded.


tariqksoliman (changes requested):
Awesome! Could you move the new code up into the above switch statement to make it specific for 'raster' layers? Also, if you could fill out a sentence description in the documentation, that'd be great.

tariqksoliman (approved):
Thanks and LGTM!

Issue #561, opened 2024-06-19
[New Feature]: A quick way to produce colormaps for many tiled layers


jtroberts:
### Checked for duplicates

Yes - I've already checked

### Alternatives considered

Yes - and alternatives don't suffice

### Describe the feature request

I made updates to the bulk_tiles tool and added a tool to easily produce colormaps for raster files. I did this a few months ago for a mission and think it might be useful for others.

PR #562, opened 2024-06-19
Improvements to bulk_tiles and added quantize_colormap #561

jtroberts:
## Purpose
- A quick way to produce colormaps for many tiled layers
## Proposed Changes
- quantize_colormaps will produce quantized colormaps using standard MPL palettes for a single tiff or directory of files
- New option to use continuous/discrete colors for bulk_tiles
- Other minor improvements to bulk_tiles
## Issues
- #561 
## Testing
- Used these tool updates to produced tiled layers for a mission

### DIFF
--- a/auxiliary/bulk_tiles/README.md
+++ b/auxiliary/bulk_tiles/README.md
@@ -11,6 +11,7 @@ optional arguments:
   -h, --help            show this help message and exit
   -c COLORMAP_DIR, --colormap_dir COLORMAP_DIR
                         Directory containing colormaps
+  -d, --discrete        Use discrete color values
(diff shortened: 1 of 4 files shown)


tariqksoliman (approved):
Awesome! LGTM

(One minor gripe is the usage of `python3` over `python` but I can see arguments for both sides/preferences so no need to change anything,)

jtroberts:
> (One minor gripe is the usage of `python3` over `python` but I can see arguments for both sides/preferences so no need to change anything,)

Yeah, I'm not sure what the best practice is these days, but I've run into a lot of systems that only use `python3` and don't include an alias to `python`.

Issue #556, opened 2024-06-14
[New Feature]: Apply scale factor to Identifier Tool source data readouts


jtroberts:
### Checked for duplicates

Yes - I've already checked

### Alternatives considered

Yes - and alternatives don't suffice

### Related problems

When using the Identifier tool, I have some layers where their data values have been scaled to make them more human readable or to match a more common unit of measurement. In the example below, the legend has a scale factor manually applied (0.01), but when the tool queries the source data value, there is no way for it to apply the same scale factor, so it just reports the raw result.

<img width="1058" alt="Screenshot 2024-06-13 at 6 00 23 PM" src="https://github.com/NASA-AMMOS/MMGIS/assets/5315956/3dcf0410-4a9d-49ba-8da3-b84866bcb1fa">

<img width="1058" alt="Screenshot 2024-06-13 at 6 00 35 PM" src="https://github.com/NASA-AMMOS/MMGIS/assets/5315956/3fc788f4-b3e8-4549-9eeb-12b504fc1ea3">

### Describe the feature request

I want a way to specify a scale factor in the layer configuration and have the Identifier tool apply it when it reports the data value.

Issue #555, opened 2024-06-14
[New Feature]: Consistent text alignment in Identifier Tool readouts


jtroberts:
### Checked for duplicates

Yes - I've already checked

### Alternatives considered

Yes - and alternatives don't suffice

### Related problems

When I use the Identifier tool to read values for a raster layer, I find it annoying when it shows the initial color-to-legend reading text-aligned to the left (screenshot 1) and then it changes the text alignment to the right when the actual value is returned (screenshot 2). For single-band data, the band number is not relevant and may add confusion as well.

<img width="1058" alt="Screenshot 2024-06-13 at 5 54 46 PM" src="https://github.com/NASA-AMMOS/MMGIS/assets/5315956/f747a6a5-2bfe-43b8-b12d-3bbfcf52d472">

<img width="1058" alt="Screenshot 2024-06-13 at 5 54 56 PM" src="https://github.com/NASA-AMMOS/MMGIS/assets/5315956/c63765d8-cc80-49a8-beca-be895cd8d94a">

### Describe the feature request

I want the Identifier tool text readouts to have consistent alignment and a smooth transition from legend value to source value.

PR #567, opened 2024-06-21
Enabled scalefactor and improved text readout for Identifier tool

jtroberts:
## Purpose
- Improvements to the Identifier tool to allow a scale factor for specified layers and better text alignment in data readouts.
## Proposed Changes
- Add support for a "scalefactor" Identifier tool variable.
- Change data readout when there is just a single band value.
## Issues
- Consistent text alignment in Identifier Tool readouts #555
- Apply scale factor to Identifier Tool source data readouts #556
## Testing
- Added scalefactor to a few layers and tested existing layers without new variable.
- Viewed text readouts to ensure data accuracy


tariqksoliman (approved):
LGTM!

Issue #568, opened 2024-06-27
[Bug]: Deeplink to selected point or polygon fails


tariqksoliman:
### Checked for duplicates

Yes - I've already checked

### Describe the bug

When selecting a point, creating a deep link and returning to it:
```
TypeError: Cannot read properties of undefined (reading 'lat')
    at Object.selectPoint (Layers_.js:2014:1)
    at Object.fullyLoaded (Layers_.js:153:1)
```

When selecting a polygon, creating a deep link and returning to it:
- Polygon is not selected/highlighted

Note: FWIW, This was on a time enabled layer.

### What did you expect?

Deeplinks should always keep selected feature reselected.

tariqksoliman:
First issue was that if a point was deeplinked to, the assumption was made that the layer contained only point features and thus failed when it encountered a non-point.

PR #569, opened 2024-06-27
#568 fix Deeplink to selected point or polygon

tariqksoliman:
- Adds better field checking for points
- Adds vector raw-variable `useKeyAsId` to help select unique features. (Point feature's were working based on lat,lng but passing all latlngs for a polygon in a url is unfeasible) `useKeyAsName` takes precedence over latlng deeplink selections,


tariqksoliman:
Closes #568

Issue #570, opened 2024-06-27
[Bug]: RawVariables Info section does not work in dynamicExtent is true


tariqksoliman:
### Checked for duplicates

Yes - I've already checked

### Describe the bug

The info raw variable pre-supposes that the geojson is complete whereas dynamicExtent queries only parts of the geojson.

### What did you expect?

Quick info regarding the latest feature should still be shown regardless of whether dynamicExtent is true or not.

PR #571, opened 2024-06-28
#570 Info with dynamicExtent

tariqksoliman:
If the `info` layer raw variable is used with `dynamicExtent`, then it will use the geodatasets `/search` endpoint to query the necessary metadata.


tariqksoliman:
Closes #570

PR #632, opened 2025-02-27
Avoid aggressively trimming white space in text boxes in configure-beta

ac-61:
This fixes the bug for the "text" input box type where it did not allow adding spaces at the end of the text block. This adds a new input type that does not trim whitespace from text boxes in configure-beta. For now, we want to allow white spaces for layer names and units.


tariqksoliman (approved):
Perfect and thank you!

PR #636, opened 2025-03-03
Add image layer type to support GeoTIFF and Cloud Optimized GeoTIFF images

ac-61:
This adds a feature to create image layers to support loading GeoTIFF and Cloud Optimized GeoTIFF images using the [georaster-layer-for-leaflet](https://github.com/GeoTIFF/georaster-layer-for-leaflet) library. Color ramps can only be applied to single banded images. There are some issues with loading multiband images (but they can be loaded).


tariqksoliman:
Things I changed:
- Symlinks don't work in Windows so the configure-beta page failed to build until I copied the contents on js-colormaps.js into its own.:
```
Failed to compile.

SyntaxError: C:\Users\tsoliman\Documents\Projects\MMGIS\configure\src\external\js-colormaps.js: Unexpected token (1:0)
> 1 | ../../../src/external/js-colormaps/js-colormaps.js
```

tariqksoliman (approved):
LGTM

jtroberts (approved):
Looks great!

PR #638, opened 2025-03-04
Fix symlink bug

ac-61:
There was a slightly weird commit in #636, where a file was still marked as a symlink, despite not being a symlink.

Current `development` branch:
Screenshot 2025-03-04 at 2 04 51 PM

This pull request:
Screenshot 2025-03-04 at 2 04 57 PM


tariqksoliman (approved):
Ah, okay. My commit to change that was bad then. Thanks for catching this. If the docker build and push succeeds again, feel free to merge.

PR #647, opened 2025-03-11
Fix bug with fill min/max when modifying color scale in image layer

ac-61:
This fixes a bug in #636, where if the "fill min/max" option was unchecked in the configuration options for image layers, the maximum value display is incorrect when adjusting the color scale values in the LayerTool.


tariqksoliman (approved):
LGTM!

Issue #596, opened 2024-11-07
[New Feature]: Visualize velocity data (e.g. wind) as particles or streamlines


jtroberts:
### Checked for duplicates

Yes - I've already checked

### Alternatives considered

Yes - and alternatives don't suffice

### Describe the feature request

There are mapping applications that visualize winds as particles/streamlines and it would be nice to have this in MMGIS to overlay with other datasets.

PR #656, opened 2025-03-21
Dynamic legend for velocity layers

jtroberts:
## Purpose
- Enable dynamic legend with units for velocity layers
## Proposed Changes
- Added configuration option to specify legend units
- Dynamic legends based on work done for COGs (similar variable names were left)
- Minor readme updates and updated nvmrc
## Issues
- #596 
## Testing
- Added velocity streamline layer
- Legend based on colormap automatically appears
- Layer settings now contain color settings to rescale min and max values


tariqksoliman:
- It looks like the titiler endpoint doesn't like uppercase colormap names:
  - This doesn't work: http://localhost:8888/titiler/colorMaps/RDYLBU_R?format=png
  - But this does http://localhost:8888/titiler/colorMaps/rdylbu_r?format=png
  - I made a new Velocity layer and in the layer config object it's also defined as `colorScale: "RDYLBU_R"`
 - Also not in this PR but it looks like in the configure page the Velocity layers and the Image layers are both using the pink color instead of that sea-foam green.

src/essence/Tools/Layers/LayersTool.js
```diff
@@ -799,6 +798,83 @@ function interfaceWithMMGIS(fromInit) {
...
+
+                    if (
+                        node[i].kind === 'streamlines'
+                    ) {
+                        if (window.mmgisglobal.WITH_TITILER === "true") {
+                            // prettier-ignore
+                            additionalSettings = [
+                                `<img id="titlerCogColormapImage_${node[i].name}" src="${window.location.origin}${(
+                                            window.location.pathname || ''
+                                        ).replace(/\/$/g, '')}/titiler/colorMaps/${node[i].variables.streamlines.colorScale}?format=png"></img>`,
```
tariqksoliman: `node[i].variables.streamlines.colorScale` should be lowercased first.

tariqksoliman (changes requested):

tariqksoliman (approved):
LGTM

PR #662, opened 2025-03-25
Fix image layer url path

ac-61:
This fixes an issue where the url path for images in the image layer was incorrect.


tariqksoliman (approved):
LGTM, thanks!

PR #667, opened 2025-03-27
Fix bug in image layer

ac-61:
This fixes a bug in #636, where a check was missing and causing an error while loading image layers that does not have an overwritten `pixelValuesToColorFn` function (i.e. layers that are multi banded).


tariqksoliman (approved):
LGTM

PR #669, opened 2025-04-02
initial add of STAC support

flynnplatt:
## Purpose
- Adds support for STAC Catalog/Collection/Item GeoJSON
## Proposed Changes
- [ADD] New vector layer prefix: `stac`, `stac-catalog`, `stac-collection`, `stac-item`
- [CHANGE] Layer config ingestion to pre-process a STAC catalog/collection/item into header/vector layers
- [ADD] Automatic parsing of STAC assets for download links


flynnplatt:
@tariqksoliman We had previously discussed using this PR as an opportunity to add a dropdown to the Vector layer config that would perform the duties of the URL prefix, however I couldn't see a clean way to implement that without fully deprecating the URL prefix and that seemed like a bigger change than necessary at this point.

tariqksoliman:
@flynnplatt Okay, no worries and no need to implement it in this PR then.

src/essence/Basics/Layers_/Layers_.js
```diff
@@ -3791,6 +3799,107 @@ function parseConfig(configData, urlOnLayers) {
         //Otherwise return 0
         return 0
     }
+
+    // recurse through a STAC layer building sublayers
+    function getSTACLayers(d) {
+        let stac_data
+        const urlPieces = d.url.split(':')
```
tariqksoliman: I had to change this line to 
```javascript
const urlPieces = d.url.split(/:(.*)/s)
```
to get it to work, otherwise it splits all the `:` and ends up only querying `https`

tariqksoliman (commented):
I tried the following vector layer URLs:
- stac:https://cmr.earthdata.nasa.gov/stac/ORNL_CLOUD/collections/AVIRIS-Classic_L1B_Radiance_2155_1/items
  - Works!
- stac:https://cmr.earthdata.nasa.gov/stac/ORNL_CLOUD/collections/AVIRIS-Classic_L1B_Radiance_2155_1
  - Gives me a header in the LayersTool but has no layers underneath it 
- stac:https://cmr.earthdata.nasa.gov/stac/ORNL_CLOUD
  - Takes a while to query and then gives me a bunch of layer headers with no other layers beneath them

src/essence/Basics/Layers_/Layers_.js
flynnplatt: good find, I'll fix that

flynnplatt: fixed in [bcf697b](https://github.com/NASA-AMMOS/MMGIS/pull/669/commits/bcf697b90dfe2dfc1cba862122f7bd1904350d83)

flynnplatt: improved in https://github.com/NASA-AMMOS/MMGIS/pull/669/commits/e856ae15112042f8353e44efcf157262aad26f3a

tariqksoliman (approved):
LGTM now!

flynnplatt:
@tariqksoliman sorry for the re-review request but I ended up changing the implementation of fetching through a STAC tree to enable parallel fetches but that meant that I had to make a string of functions `async`. I think I did it right but I'm not sure if you have a good regression test to double-check that I didn't introduce some bug in the initial load.

The change I made makes using `stac:https://cmr.earthdata.nasa.gov/stac/ORNL_CLOUD` much faster to load

tariqksoliman (approved):
LGTM and don't see any issue with parseConfig being await/asynced

Thanks!

PR #682, opened 2025-04-22
Patch #669 and #680 Merge For STAC Prefix

flynnplatt:
## Purpose
- #669 introduced parsing URL prefix during config ingestion and #680 introduced a source_type -> URL prefix conversion during config ingestion, while the code changes didn't conflict, they were actually in the wrong order of events. This fixes that.
## Proposed Changes
- [FIX] `expandLayers()` to convert source_type to a URL prefix before then parsing that prefix


tariqksoliman (approved):
Good catch!

PR #700, opened 2025-05-28
Fix bug with noDataValue for single banded COGs

ac-61:
## Purpose
- This fixes a bug where the noDataValue was not being respected when the noDataValue is in between the min/max data range on an image layer.

## Testing
Tested on a single banded COG image where the noDataValue is in between the min/max data range with the following:
- "Transform COG" checked and "Hide No Data Value" checked
- "Transform COG" checked and "Hide No Data Value" unchecked
- "Transform COG" unchecked and "Hide No Data Value" checked
- "Transform COG" unchecked and "Hide No Data Value" unchecked

### DIFF
--- a/src/essence/Basics/Map_/Map_.js
+++ b/src/essence/Basics/Map_/Map_.js
@@ -1528,6 +1528,11 @@ function makeImageLayer(layerObj) {
                 'variables.image'
             )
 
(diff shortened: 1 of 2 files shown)


tariqksoliman (approved):
LGTM!

PR #726, opened 2025-08-07
Add feature to set Layer header expanded state individually

ac-61:
This adds a feature to set the expanded/unexpanded state of each header in the Layers individually by default when opening the layers tool.
Screenshot 2025-08-07 at 11 17 05 AM_

The reset button sets all headers to its default expanded/unexpanded state from the MMGIS configuration accordingly. 
Screenshot 2025-08-07 at 11 09 03 AM

If the "Expanded" option for Layers is set the "Tools" configuration, opening the Layers tool has the original behavior, where all the headers start in the expanded state.
Screenshot 2025-08-07 at 11 19 01 AM_

### DIFF
--- a/configure/src/metaconfigs/layer-header-config.json
+++ b/configure/src/metaconfigs/layer-header-config.json
@@ -30,6 +30,14 @@
               "description": "",
               "type": "textnotrim",
               "width": 8
+            },
+            {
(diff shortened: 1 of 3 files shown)


tariqksoliman (approved):
Nice! LGTM and thank you!

Issue #652, opened 2025-03-18
[New Feature]: Support more platforms for Docker builds


fcalef:
### Checked for duplicates

Yes - I've already checked

### Alternatives considered

Yes - and alternatives don't suffice

### Related problems

Deploying MMGIS via Docker makes the installation process easier than a manual install. However, current MMGIS Docker builds only support linux platforms.

### Describe the feature request

I want to have additional platforms supported in the Docker builds, e.g. Rasberry PI, Mac M1/M2, Windows 11.

PR #733, opened 2025-08-19
Include amd64 Docker image suffix and build it last

jtroberts:
## Purpose
- Include amd64 Docker image suffix and build it last so that it takes precedent over arm64.
- Remove extraneous image tags and unknown OS when building Docker images with GitHub Actions.
## Proposed Changes
- Add -amd64 to the end of "default" Docker image names
- Build amd64 after arm64 to ensure it's the latest
- Revert to using Docker build instead of buildx to prevent extraneous image tags
## Issues
- #652 
## Testing
- Testing builds with PR


tariqksoliman (approved):
Approving but I do still see `unknown/unknown` in https://github.com/NASA-AMMOS/MMGIS/pkgs/container/mmgis and -arm64 for both

PR #734, opened 2025-08-25
Filter out blank csv entries in csvToJSON function

ac-61:
It seems it's possible to read a csv file and get a blank entry at the end. We should probably filter those out.

Screenshot 2025-08-25 at 1 44 46 PM


tariqksoliman (approved):
LGTM

PR #735, opened 2025-08-27
Add Legend tool display options

ac-61:
This adds some additional options to the Legend tool.

Each layer that can display legends has an option to turn off the layer name in the Legend tool, which is useful if a layer has only a single legend item that has the same name as the layer name.

Screenshot 2025-08-27 at 2 42 32 PM
Screenshot 2025-08-27 at 2 53 40 PM

The Legend tool control has an option to show "Header" type layer names in the Legend tool. This will automatically display the child layers in a nested view.

Screenshot 2025-08-27 at 2 41 46 PM

mmgis_legends_01


tariqksoliman (approved):
Looks great; thank you!

PR #759, opened 2025-09-18
Show color scale even if TiTiler PNG images are not available

ac-61:
If the TiTiler server is somehow down, but TiTiler is on according to the env, the color scales will display an error. We should fall back to use js-colormaps in this case for both the layers tab and the configure page.

Screenshot 2025-09-12 at 1 01 33 PM

Screenshot 2025-09-15 at 11 08 35 AM

Screenshot 2025-09-18 at 1 23 18 PM

Screenshot 2025-09-18 at 1 26 51 PM

### DIFF
--- a/configure/src/core/Maker.js
+++ b/configure/src/core/Maker.js
@@ -269,6 +269,113 @@ const useStyles = makeStyles((theme) => ({
   },
 }));
 
+const ColormapHelper = ({ colormapName }) => {
+  const c = useStyles();
+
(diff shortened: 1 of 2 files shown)


tariqksoliman (approved):
Approving but I do also think if TiTiler is down when it's supposed to be up, that there's a larger issue with the deployment that should be addressed instead.

jtroberts (approved):

Issue #756, opened 2025-09-16
[New Feature]: Enable horizontal legends


jtroberts:
### Checked for duplicates

Yes - I've already checked

### Alternatives considered

Yes - and alternatives don't suffice

### Related problems

We're currently displaying vertical legends (unless using a custom graphic) in the legend tool. It would be nice to also support horizontal legends to save on vertical screen space.

### Describe the feature request

Enabled an option to display a legend either horizontally or vertically.

PR #760, opened 2025-09-23
Enable horizontal legends

jtroberts:
## Purpose
- Enable an option to display legends horizontally in addition to the default of vertically.
## Proposed Changes
- New configuration option to select horizontal or vertical legend
- Display continuous or discrete legends with ticks and appropriate label sizes (will condense labels if too many for display)
- Condense units and display at the top of the legend to avoid label cluttering
- Show tooltips of legend values when hovering over the legends (any type of legend including vertical)
## Issues
- (https://github.com/NASA-AMMOS/MMGIS/issues/756)
## Testing
- Tested with various legend types and ensured existing legends work properly.
- Examples:

Screenshot 2025-09-22 at 7 03 54 PM
<p>
Screenshot 2025-09-22 at 6 57 26 PM

### DIFF
--- a/configure/src/metaconfigs/layer-image-config.json
+++ b/configure/src/metaconfigs/layer-image-config.json
@@ -199,7 +199,16 @@
               "name": "Legend From URL",
               "description": "A URL to a static image or .csv with the following header: 'color,strokecolor,shape,value'. If the path is relative, it will be relative to the mission's directory. This legend is overridden if a legend is also configured below.",
               "type": "text",
-              "width": 10
+              "width": 8
+            },
+            {
+              "new": true,
+              "field": "variables.legendOrientation",
+              "name": "Legend Orientation",
+              "description": "Displays the legend horizontally or vertically.",
+              "type": "dropdown",
+              "width": 2,
+              "options": ["vertical", "horizontal"]
             },
             {
               "new": true,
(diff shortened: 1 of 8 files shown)


tariqksoliman (changes requested):
LGTM!

Only two things of note I found:
- When the legend orientation is horizontal, the definitions for continuous legends seem inverted than what I'd expect. 0 -> 100 (top -> bottom) in the legend configuration but 100 -> 0 in the legend tool. 
- The hover over text seems to be relying entirely on the legends "Label" configuration for continuous legends. I think it should prioritize "Property Value" else "Label".

tariqksoliman:
Maybe the first bullet point isn't a problem -- it does make sense to have the higher values on top.

jtroberts:
> Maybe the first bullet point isn't a problem -- it does make sense to have the higher values on top.

That makes more sense to me so I left it as-is. 

> * The hover over text seems to be relying entirely on the legends "Label" configuration for continuous legends. I think it should prioritize "Property Value" else "Label".

I've updated the hover behavior to show the "raw" property value.

Also fixed an issue I found where the units would display only for continuous legends.

tariqksoliman (approved):
Thank you and LGTM now!

PR #767, opened 2025-10-02
Adds frozon API proxy

jdrodjpl:
...at `/frozon_api`


tariqksoliman (approved):
I also made it so that adjacent server proxying can be set up fully through envs like:
```
ADJACENT_SERVER_CUSTOM_{X}=["true", "my_api", "my_api_service", "8105"]
```

PR #805, opened 2025-11-13
Fixes in TimeUI for mobile mode

ac-61:
This fixes an issue where the user selected time did not persist when the TimeUI is opened and closed in mobile mode. This also shifts to view the selected elements in the expanded timeline every time the user selects a new time.


tariqksoliman (approved):
LGTM and thanks!

PR #811, opened 2025-11-19
Fix scale bar position in mobile mode

ac-61:
The scale bar was partially hidden under the toolbar.

Screenshot 2025-11-19 at 3 11 59 PMScreenshot 2025-11-19 at 3 11 38 PM


tariqksoliman (approved):
Thanks!

PR #816, opened 2025-12-01
Add new callback for newActiveFeature

ac-61:
This adds a new event listener callback,`newActiveFeature`, which is triggered when a feature is clicked.


tariqksoliman (approved):
LGTM!

PR #817, opened 2025-12-03
Highlight buttons in extended timeline view as range indicator in mobile mode

ac-61:
Since we removed the yellow range indicators in mobile mode for the expanded timeline, highlight the buttons instead when in mobile mode.

Screenshot 2025-12-03 at 11 10 25 AM


tariqksoliman (approved):
LGTM!

PR #819, opened 2025-12-03
Fix issue with local versus UTC time in expanded timeline display

ac-61:
This fixes an issue where there was a local time versus UTC time offset in the expanded timeline display.

Before (End time set to 01/01/2025)
Screenshot 2025-12-03 at 12 52 01 PM
Before (End time set to 01/02/2025)
Screenshot 2025-12-03 at 12 52 06 PM

After (End time set to 01/01/2025)
Screenshot 2025-12-03 at 12 52 28 PM
After (End time set to 01/02/2025)
Screenshot 2025-12-03 at 12 52 38 PM

### DIFF
--- a/src/essence/Ancillary/TimeUI.js
+++ b/src/essence/Ancillary/TimeUI.js
@@ -1878,156 +1878,17 @@ const TimeUI = {
         // Populate Hours Row
         TimeUI._populateHoursRow()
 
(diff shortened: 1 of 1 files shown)


tariqksoliman (approved):
Nice and thank you!

PR #846, opened 2026-01-08
Add new callback for layersToolHeaderStateChange

ac-61:
## Purpose
This adds a new event callback `layersToolHeaderStateChange`, which returns the expanded state of a header that is clicked. This is useful because we can use it to keep track of the headers that are expanded and display the same header state when opening the LayersTool.


tariqksoliman (approved):
LGTM!

PR #848, opened 2026-01-13
Fix LegendTool overflow

ac-61:
## Purpose
This makes sure the LegendTool items are attached to the correct container, otherwise there is no scrollbar.

Before -> After
Screenshot 2026-01-13 at 9 36 58 AM Screenshot 2026-01-13 at 9 37 08 AM


tariqksoliman (approved):
Thank you!

PR #855, opened 2026-01-28
Add viewer_open as new Kind

ac-61:
This adds `viewer_open` as a new Kind. This will zoom to a clicked feature and expand the Viewer panel if it is not already open.

Screenshot 2026-01-28 at 2 51 07 PM

This works on GeoJSON features with the following geometry types: Point, LineString, Polygon, MultiPoint, MultiLineString, and MultiPolygon.

For points, the map will zoom to the zoom level set in "Zoom Level of Map Scale" (found under the UI configuration), if it is set, otherwise it will use the current zoom level of the map. For all other geometry types, the map will zoom to the bounds of the feature.

mmgis_kinds_01


tariqksoliman (changes requested):
Looks good but I'm seeing that when selecting a feature with the viewer_open kind, if I pan or zoom away from the feature while it's still selected, it automatically snaps back to the feature. Could you make it so that it doesn't keep fixating on the features after the initial selection. Thanks!

tariqksoliman (approved):
I couldn't replicate the issue I found so I'm approving.

PR #857, opened 2026-01-30
#856 AnimationTool Improvements

### LINKED ISSUE #856: [Bug]: AnimationTool outputs is warped in non-mercator projections
Also include any other AnimationTool touchups

tariqksoliman:
Closes #856

_With Claude_

- Fixes warped outputs on non-mercator projections
- Now animates in an offscreen replica map
- Includes ffmpeg scripts so that mp4 downloads aren't blocked by CSP
- Various UI touchups

Screenshot 2026-01-29 193541


jtroberts (approved):
Looks good to me! A few minor things of note that could be improved:

1. Velocity type layers don't render in the animation
2. Scale bar and title text is sometimes too small depending on area selected and screen resolution
3. Include "Z" at the end of time label

Including the legend in the animation would also be nice but probably should be its own issue.

tariqksoliman:
Addressed:
2. Scale bar and title text is sometimes too small depending on area selected and screen resolution
3. Include "Z" at the end of time label
In my latest commit and also including the ability to cancel still-generating animations.

PR #858, opened 2026-02-03
Add new callback for madeLegendTool

ac-61:
This adds a callback for `madeLegendTool`, which sends an event whenever the LegendTool's `make` function is called. This is useful for overriding what is displayed in the LegendTool by default.


tariqksoliman (approved):
LGTM

PR #862, opened 2026-02-11
Fix return value for layersToolHeaderStateChange event

ac-61:
Fixes the return value for the event in #846.


tariqksoliman (approved):
LGTM

PR #865, opened 2026-02-17
Fix bug in viewer_open kind

ac-61:
This fixes a bug in #855 where the incorrect lat/long property was being used.


tariqksoliman (approved):
Thanks!

Issue #871, opened 2026-02-25
[Bug]: DyanmicExtent+Threshold Layers do not properly update


tariqksoliman:
DyanmicExtent+Threshold Layers do not properly update if underlying the data changes, but the user never pans or zooms the map.

To test:
Configure a vector layer to have:
- Refresh Interval Enabled: true
- Dynamic Extent: true
- Dynamic Extent Threshold: 200000/z

Manually update the underlying vector json file and observe how, despite refreshing through time, the layer on the map does not update until the user pans.

The layer should be updated properly when using the JavaScript API functions as well.

PR #873, opened 2026-02-26
#871 Bug: DyanmicExtent+Threshold Layers do not properly update

tariqksoliman:
Address #871


tariqksoliman:
Was done in MMGIS-2.12.3 instead

PR #874, opened 2026-02-26
Add font types as asset to webpack

ac-61:
This is an attempt to fix an issue where the material design icons all show up as boxes because the fonts seem to be missing. We're not able to consistently replicate this bug.

Screenshot 2026-02-26 at 9 19 24 AM


tariqksoliman (approved):
LGTM

PR #875, opened 2026-02-26
Fix issues in mobile version

ac-61:
This fixes a few issues with the mobile version.

This fixes a toolbar bug where if clicking the LayerTool, InfoTool, or LegendTool, then clicking the TimeUI or Coordinates tool and then clicking the previous tool again, it would turn off the previous tool instead of switching to it.

This changes the height of the  LayerTool, InfoTool, and, LegendTool to be dynamic based on percentage of the height of the MMGIS page.

Screenshot 2026-02-26 at 12 37 45 PM_ Screenshot 2026-02-26 at 12 37 59 PM Screenshot 2026-02-26 at 12 37 56 PM_

Larger toolbar buttons before -> after

Screenshot 2026-02-26 at 12 27 34 PM
Screenshot 2026-02-26 at 12 27 10 PM

### DIFF
--- a/configure/package.json
+++ b/configure/package.json
@@ -1,6 +1,6 @@
 {
   "name": "configure",
-  "version": "4.2.12-20260226",
+  "version": "4.2.13-20260226",
   "homepage": "./configure/build",
   "private": true,
   "dependencies": {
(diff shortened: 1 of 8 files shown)


tariqksoliman (approved):
LGTM, Thanks!

PR #882, opened 2026-03-04
Fix bug in viewer_open kind

ac-61:
This makes sure to check to see if the viewer panel is enabled/exists before opening it.


tariqksoliman:
@ac-61 Why `L_.UserInterface_.openViewerPanel` instead of `L_.hasViewer`?

ac-61:
@tariqksoliman No one saw that 🙈

tariqksoliman (approved):
Ha LGTM!

PR #883, opened 2026-03-04
Improve login page for smaller screens

### LINKED ISSUE #877: [New Feature]: Improve login page for mobile version
### Checked for duplicates

Yes - I've already checked

### Alternatives considered

Yes - and alternatives don't suffice

### Related problems

The elements on the login page in mobile mode is very small. The font size is also less than 14px so it causes mobile browsers to zoom in on the input boxes.

Image

### Describe the feature request

We should make a mobile version of the login page or fix the desktop version to be adaptable for both desktop and mobile.

ac-61:
This closes #877. This should work well for screens that are at least 330 pixels wide. Some improvements could be made for screens that are less than 330 pixels wide.

Before -> After
Screenshot 2026-03-04 at 2 16 20 PM Screenshot 2026-03-04 at 2 33 48 PM

### DIFF
--- a/configure/package.json
+++ b/configure/package.json
@@ -1,6 +1,6 @@
 {
   "name": "configure",
-  "version": "4.2.17-20260305",
+  "version": "4.2.18-20260305",
   "homepage": "./configure/build",
   "private": true,
   "dependencies": {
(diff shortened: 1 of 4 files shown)


tariqksoliman (approved):
Nice! Thank you.

PR #899, opened 2026-03-18
Fix image loading in OpenSeadragon

ac-61:
This fixes an issue when loading images with OpenSeadragon in the Viewer panel. The images were always loaded at a zoom level of 50%. This changes it so the images are always loaded at 100% and images should now stretch to fit the size of the Viewer panel when it is resized.


tariqksoliman (approved):
Approving (but it would be better if it was possible that the image fitted to fill which ever x/y extent was the smallest. So, in the image below, the companion cube below would fill the height and not the width.)

Image

ac-61:
@tariqksoliman Should be fixed 😁 
Screenshot 2026-03-19 at 11 44 01 AM

tariqksoliman (approved):
Nice LGTM!

PR #905, opened 2026-03-19
Add option to display Latitude/Longitude in coordinates display

ac-61:
This adds an option in the Coordinates tab in the configure page to display coordinates as "Latitude, Longitude" instead. 

Screenshot 2026-03-19 at 3 53 03 PM


tariqksoliman (approved):
LGTM, Thanks!

PR #923, opened 2026-04-06
Add option to set tool to open by default

ac-61:
This adds an option in the configure page's User Interface tab to allow users to set a single tool to be opened by default when starting MMGIS. This only allows tools that are not "separated tools" to be open by default.


tariqksoliman (changes requested):
When I first toggled the Default Tool to be enabled, it didn't open the LayersTool be default. I had to select a different tool in the dropdown and then Layers again for the change to take effect. I know this is an issue with the dropdown component. I think the dropdown components within the layer modals do automatically default to their listed default but it looks like that's not true for dropdown components that are only in the Tab pages. 

An easy fix without fixing the dropdowns in the Tabs would be to change the Default Tool section from a toggle and dropdown to just a dropdown that defaults to "None" (Thus forcing a change in the dropdown component so that it registers when being saved)

tariqksoliman (approved):
LGTM! Thanks!

PR #942, opened 2026-04-21
  feat: Add 3D extrusion for vectortile layers + 3D Tiles support

jdrodjpl:
Adds two new 3D rendering capabilities to the Cesium globe renderer:

  1. Vector tile extrusion: vectortile layers can now render extruded 3D buildings on the globe via a new "3D Extrusion" tab in the layer config. A CesiumMVTLayer class manages tile lifecycle (load, decode, evict) with batched Cesium.Primitive rendering and per-feature color support from the OpenMapTiles schema. No Cesium ion token required — works with any MVT source (OpenFreeMap, Versatiles, self-hosted).

  2. 3D Tiles layer type: new "3dtiles" layer type for Cesium3DTileset URLs, with configurable LOD, memory limits, height offset, and style expressions.

  Also:

  - New auxiliary/resolve-tile-url CLI utility to resolve TileJSON endpoints to concrete tile URLs (keeps MMGIS provider-agnostic).
  - Scene lighting enabled with a fixed sun angle (summer solstice, 10am EDT) for consistent, readable building shading.
  - L.vectorGrid sublayer filtering in Map_.js to hide MVT sublayers not explicitly styled (prevents default blue rendering of roads, water, etc.).
  - Backend validation (API/Backend/Config/validate.js) accepts the new 3dtiles type.


tariqksoliman (changes requested):
Very cool and love the addition to the Reference-Mission. A couple of suggestions:

- The rendering in the 2D map is a bit slow and it may be worthwhile running the building footprints through some simplification algorithm first such as https://turfjs.org/docs/api/simplify
- The buildings in 3D do not sit on the terrain. They look to all be at an elevation of 0.

jdrodjpl:
Added fix to the elevations issue and using a simplification routine to break down the geometry a little bit, seems faster to me.

With cities particularly high in elevation like Denver, there's an issue where the user has to zoom in really close to see buildings render. That's because the buildings really only appear in a high zoom level, and that zoom level (i.e., camera above earth's surface) is much closer to Denver than it is San Francisco.

tariqksoliman:
I'm still seeing the following regarding the elevation:
extruded_osm_buildings_elev

jdrodjpl:
Seems better now: 
image

tariqksoliman (approved):
Nice! LGTM!

PR #989, opened 2026-05-28
fix: correct adjacent server URL rewriting when behind a reverse proxy/load balancer

### LINKED ISSUE #753: [Bug]: Adjacent servers using internal Docker container name instead of server name in URLs
### Checked for duplicates

Yes - I've already checked

### Describe the bug

I'm running MMGIS in Docker with stac-fastapi-pgstac. I'm having trouble with stac-fastapi respecting the server name

If I hit the STAC catalog with just the IP address, the stac collection uses that IP as the server name in its URLs. 

If I use the load balancer address, it uses "http://stac-fastapi/" instead of the server name.

### What did you expect?

I expect URLs in STAC results to match the server and path used to make the request. 

### Reproducible steps

```bash

```

### Environment

```bash

```

sujen1412:
## Summary

Fixes #753

When MMGIS is deployed behind a load balancer that terminates SSL (e.g. AWS ALB), adjacent services like `stac-fastapi` receive proxied requests where `changeOrigin: true` rewrites the `Host` header to the internal Docker service name (e.g. `stac-fastapi`). The upstream service then builds self-referencing URLs using that internal hostname.

The existing `replaceAll` in `createSwaggerInterceptor` only matched the explicit `http://service:port` target string — but when the upstream service also picks up `X-Forwarded-Proto: https` from the ALB (passed through unchanged by the proxy), it generates URLs like `https://stac-fastapi/collections` instead of `http://stac-fastapi:8881/collections`. This variant was never matched, so internal hostnames leaked through to clients.

## Changes

- **`proxyReq` handler on the STAC proxy**: forwards `X-Forwarded-Host` (the real public hostname) and `X-Forwarded-Proto` to the upstream service, so it can natively build correct public URLs when configured with `--proxy-headers`
- **Broadened URL replacement in `createSwaggerInterceptor`**: extracts the service hostname from the target URL and replaces all variants (`http://`, `https://`, with and without port) as a reliable fallback for any upstream service

## Testing

Verified on a deployment behind an AWS ALB with `stac-fastapi-pgstac`. Before the fix, STAC catalog responses contained `https://stac-fastapi/collections`. After the fix, responses correctly contain `https://<public-domain>/stac/collections`.

For `stac-fastapi`, also add the following flags to the uvicorn command so it trusts the forwarded headers:
```
--proxy-headers --forwarded-allow-ips='*' --root-path /stac
```

### DIFF
--- a/adjacent-servers/adjacent-servers-proxy.js
+++ b/adjacent-servers/adjacent-servers-proxy.js
@@ -25,6 +25,10 @@ function initAdjacentServersProxy(app, isDocker, ensureAdmin, ensureUserForAdjac
         },
         selfHandleResponse: true,
         on: {
+          proxyReq: (proxyReq, req, res) => {
+            proxyReq.setHeader("X-Forwarded-Host", req.get("host"));
+            proxyReq.setHeader("X-Forwarded-Proto", req.headers["x-forwarded-proto"] || req.protocol);
+          },
           proxyRes: createSwaggerInterceptor("stac", stacTarget),
         },
       })
@@ -272,6 +276,10 @@ const createSwaggerInterceptor = (path, target) => {
         res.get("Content-Type").includes("html"))
     ) {
       newResponse = newResponse || responseBuffer.toString("utf8");
+      const serviceHost = new URL(target).hostname;
+      newResponse = newResponse.replaceAll(`https://${serviceHost}/`, `${req.protocol}://${req.get("host")}/${path}/`);
+      newResponse = newResponse.replaceAll(`https://${serviceHost}`, `${req.protocol}://${req.get("host")}/${path}`);
+      newResponse = newResponse.replaceAll(`http://${serviceHost}/`, `${req.protocol}://${req.get("host")}/${path}/`);
       newResponse = newResponse.replaceAll(
         target,
         `${req.protocol}://${req.get("host")}/${path}`
--- a/adjacent-servers/adjacent-servers-proxy.js
+++ b/adjacent-servers/adjacent-servers-proxy.js
@@ -51,6 +51,10 @@ function initAdjacentServersProxy(app, isDocker, ensureAdmin, ensureUserForAdjac
         },
         selfHandleResponse: true,
         on: {
+          proxyReq: (proxyReq, req, res) => {
+            proxyReq.setHeader("X-Forwarded-Host", req.get("host"));
+            proxyReq.setHeader("X-Forwarded-Proto", req.headers["x-forwarded-proto"] || req.protocol);
+          },
           proxyRes: createSwaggerInterceptor("tipg", tipgTarget),
         },
       })
@@ -89,6 +93,10 @@ function initAdjacentServersProxy(app, isDocker, ensureAdmin, ensureUserForAdjac
         },
         selfHandleResponse: true,
         on: {
+          proxyReq: (proxyReq, req, res) => {
+            proxyReq.setHeader("X-Forwarded-Host", req.get("host"));
+            proxyReq.setHeader("X-Forwarded-Proto", req.headers["x-forwarded-proto"] || req.protocol);
+          },
           proxyRes: createSwaggerInterceptor("titiler", titilerTarget),
         },
       })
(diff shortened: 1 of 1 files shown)


adjacent-servers/adjacent-servers-proxy.js
```diff
@@ -272,6 +276,10 @@ const createSwaggerInterceptor = (path, target) => {
         res.get("Content-Type").includes("html"))
     ) {
       newResponse = newResponse || responseBuffer.toString("utf8");
+      const serviceHost = new URL(target).hostname;
+      newResponse = newResponse.replaceAll(`https://${serviceHost}/`, `${req.protocol}://${req.get("host")}/${path}/`);
+      newResponse = newResponse.replaceAll(`https://${serviceHost}`, `${req.protocol}://${req.get("host")}/${path}`);
```
tariqksoliman: _With Devin_

### Broadened URL replacement in createSwaggerInterceptor — Has a bug
The new replacements extract the hostname and do:

```javascript
const serviceHost = new URL(target).hostname;  
newResponse = newResponse.replaceAll(`https://${serviceHost}/`, `.../${path}/`);  
newResponse = newResponse.replaceAll(`https://${serviceHost}`,  `.../${path}`);  
newResponse = newResponse.replaceAll(`http://${serviceHost}/`,  `.../${path}/`);  
// then the existing target replacement
```
The problem: the pattern https://${serviceHost} (without trailing slash) will greedily match inside URLs that contain the hostname followed by a port. For example, if a response contains https://stac-fastapi:8881/foo, step 2 matches the https://stac-fastapi prefix and produces:

```
https://public-host/stac:8881/foo
```
...which is broken. The same issue applies to the http://${serviceHost} variant (though that one is partially saved by the existing target replacement running after, it still corrupts if the protocol differs).

Suggested fix: Either:

- Use a regex with a negative lookahead to avoid matching when followed by :/port: /https?:\/\/${serviceHost}(?!:\d)/
- Or add explicit https://${serviceHost}:${port} replacements before the bare-hostname replacements, so the more-specific patterns consume those occurrences first

adjacent-servers/adjacent-servers-proxy.js
```diff
@@ -25,6 +25,10 @@ function initAdjacentServersProxy(app, isDocker, ensureAdmin, ensureUserForAdjac
         },
         selfHandleResponse: true,
         on: {
+          proxyReq: (proxyReq, req, res) => {
```
tariqksoliman, line 28: Could you also include this url rewriting for the tipg, titiler and titilerpgstac proxies too. Thanks

tariqksoliman (changes requested):

tariqksoliman (changes requested):
1. **The greedy-match bug isn't fully fixed, just narrowed.** `replaceAll("https://" + serviceHost)` still corrupts any occurrence with a port other than the target's, e.g. `https://stac-fastapi:8080/x` → `https://public/stac:8080/x`. A single regex with a negative lookahead is strictly safer and shorter than the 6 sequential `replaceAll`s:
```js
new RegExp(`https?://${escapeRegExp(serviceHost)}(:${servicePort})?`, "g")  // or (?::\d+)?
```
2. **Asymmetric http/https handling.** There's `http://host/` but no bare `http://host` (no trailing slash, no port) — only the `target` string covers http, and only with the exact port. The regex above removes this whole class of gaps.
3. **The two halves of the fix can conflict.** If a deployment follows the PR's advice and enables `--proxy-headers` without `--root-path /stac`, the upstream now emits `https://<public-host>/collections` — correct host, missing the `/stac` prefix, and the interceptor's fallback no longer matches (hostname isn't the service host), so it silently ships broken URLs. Either forward the headers *and* set `--root-path` in `docker-compose.sample.yml`/`docker-compose.v1-sample.yml` (the PR documents the uvicorn flags but doesn't actually change the sample compose files), or drop the header forwarding and rely purely on the rewrite.
4. **4x duplicated `proxyReq` handler** — factor out one `forwardProxyHeaders` fn. Also consider `req.headers["x-forwarded-host"] || req.get("host")` to mirror how proto is handled.
5.  No test coverage for the rewrite; it's a pure string function and easy to unit-test with the port/protocol variants above.

PR #1000, opened 2026-06-18
fix: db volume mount path for postgres 18

Originaljsx:
postgis:18 stores data in a version specific subdir and expects the volume at /var/lib/postgresql, not /var/lib/postgresql/data. On a fresh install the old path makes the db container crash-loop, so mmgis never starts (container mmgis-db-1 is unhealthy).

Left over from the postgres16 to 18 upgrade (#935). 
Tested with down -v then up -d on a clean volume.

### DIFF
--- a/docker-compose.sample.yml
+++ b/docker-compose.sample.yml
@@ -245,6 +245,6 @@ services:
       timeout: 8s
     restart: on-failure
     volumes:
-      - mmgis-db:/var/lib/postgresql/data
+      - mmgis-db:/var/lib/postgresql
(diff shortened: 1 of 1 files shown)


tariqksoliman:
Hi and nice catch! I think because I hadn't tested it on a clean volume that I missed it. So it could be an issue with existing databases living at /data and migrating to v18 but, being in the sample docker-compose, I think that's completely fine and especially since we're already pointing to a v18 postgis image in it.

Thanks again!

PR #1005, opened 2026-06-22
Legend fixes

mohatjpl:
## Purpose
Fixes several rendering issues in the Legend tool's horizontal  legends, generalizes the separated-tool panel auto-sizing so any seperatable tool avoids overlapping the map's bottom-left controls, and adds example layers to the Earth reference mission to make testing the legend easier for everyone.

## Proposed Changes
- [ADD] "Legend Source Types" section in the Earth reference-mission blueprint with an example layer for each legend source type including CSV file, inline JSON, image, and TiTiler color ramp.
- [CHANGE] Generalized the separated-tool panel max-height: it is now derived from the shared `#toolcontroller_sep_content` container and applied to every panel, so any separated tool shrinks/scrolls to clear the bottom-left compass and scale bar. Previously this was Legend-only.
- [CHANGE] Legend units now render as a chip in the layer's title row instead of floating above the bar.
- [CHANGE] Legend images scale to fit the panel width.
- [CHANGE] Legend label font set to 14px Roboto to match the rest of MMGIS.
- [FIX] Horizontal continuous tick labels are centered on their ticks with reserved bar-end padding, so the first and last labels no longer crowd together or clip at the panel edge.
- [FIX] Units chip no longer overlaps the layer title.
- [FIX] Image legends no longer overflow and clip the panel.
- [FIX] The hover-value tooltip is clamped within the bar so it is no longer cut off at the panel edge.

## Testing
- Manually tested in the running app.
- Added example layers for each legend source (CSV, inline JSON, image, and TiTiler color ramp) to the Earth reference mission and confirmed each legend renders correctly: even tick spacing, units chip placement, image scaling, and tooltip clamping, with the timeline both collapsed and expanded.


src/essence/Basics/UserInterface_/components/BottomElementPositioner/BottomElementPositioner.jsx
```diff
@@ -140,8 +143,25 @@ function BottomElementPositioner() {
                 sepContent.style.transition = 'left 0.2s ease-out'
                 sepContent.style.left = (12 + tpShift + (tpShift > 0 ? 12 : 0)) + 'px'
             }
+
+            // Cap the Legend panel's height so it never overlaps the bottom-left
+            // compass + scale bar. That stack's top sits ~70px above the bottom
+            // bar (mapToolBar bottom = totalOffset), so the legend's bottom must
+            // clear totalOffset + 70 + a 12px gap. totalOffset already grows with
+            // the expanded timeline, so the legend shrinks exactly as needed.
+            const legendEl = document.getElementById('LegendTool')
```
tariqksoliman: This logic isn't quite working right (I can see a tall legend overlapping the compass)

Also this would be awesome if it didn't just work for the legend, but for any separated Tool more broadly.

plugins/core/tools/Legend/LegendTool.js
```diff
@@ -583,62 +586,66 @@ function drawLegends(tools, _legend, layerUUID, display_name, opacity, shift) {
...
+
+        // Fixed font size for horizontal labels — no per-legend shrinking.
+        const HORIZONTAL_LABEL_FONT_PX = 12
+
+        // Thin labels to an evenly-spaced subset that fits without overlapping.
+        if (orientation === 'horizontal' && visibleLabels.length > 2) {
+            let measureText
+            try {
+                const mctx = document.createElement('canvas').getContext('2d')
+                mctx.font = `${HORIZONTAL_LABEL_FONT_PX}px sans-serif`
```
tariqksoliman: The actual font we're using in the legend currently is "14px roboto".

tariqksoliman (changes requested):
Please, for testing, also add a section of layers to the earth reference mission blueprint that are configured for each possible legend: csv, json, from an image, from titiler colorramp.

tariqksoliman:
Here's are some things I thought relevant that AI found too:

1. Race condition on legendContainer.width() — measuredWidth is read via jQuery .width() right after appending the container. If the browser hasn't laid out yet (no reflow forced), this could return 0, falling back to the hardcoded 240px. A requestAnimationFrame or explicit reflow would be safer.

2. activeSeparatedTools added to deps but unused in the effect body — it's added to trigger a re-run when Legend opens/closes (so getElementById('LegendTool') finds the element), which is clever, but a comment explaining the intentional dep-for-side-effect would prevent someone from removing it as "unused" during cleanup.

plugins/core/tools/Legend/LegendTool.js
mohatjpl: Good catch, I updated it to Roboto and 14px

src/essence/Basics/UserInterface_/components/BottomElementPositioner/BottomElementPositioner.jsx
mohatjpl: This should be working as intended now and logic extended for any separated tool.

blueprints/Missions/Reference-Mission/config.reference-mission.json
```diff
@@ -3468,6 +3468,140 @@
       "demparser": "",
       "controlled": false,
       "uuid": "3acb87e4-2097-41c9-a186-4d4ca61bb143"
+    },
+    {
+      "name": "Legend Source Types",
```
tariqksoliman: LGTM to me overall. One minor thing though: could you move the "Legend Source Types" and its layers to live under "Vector Layers" -> "Layer Configuration" -> "Legends Tab"

tariqksoliman (changes requested):

blueprints/Missions/Reference-Mission/config.reference-mission.json
mohatjpl: This should be addressed now!

src/essence/Basics/TimeControl_/TimeUI.js
```diff
@@ -38,10 +38,6 @@ const MS = {
     decade: 315576000000,
 }
 
-// Indicator lines that tools can add/remove on the timeline
-// Map<string, {id, groupId, color, time (ms timestamp)}>
-const _indicators = new Map()
```
tariqksoliman, line 43: How come these TimeUI indicator functions were removed?

plugins/core/tools/Legend/LegendTool.js
```diff
@@ -1,8 +1,10 @@
...
+
 import $ from 'jquery'
-import L_ from '@basics/Layers_/Layers_'
-import Map_ from '@basics/Map_/Map_'
-import ToolController_ from '@basics/ToolController_/ToolController_'
-import Help from '@basics/UserInterface_/components/Help/Help'
+import L_ from '../../../../src/essence/Basics/Layers_/Layers_'
+import Map_ from '../../../../src/essence/Basics/Map_/Map_'
+import ToolController_ from '../../../../src/essence/Basics/ToolController_/ToolController_'
+import Help from '../../../../src/essence/Basics/UserInterface_/components/Help/Help'
```
tariqksoliman: Can we go back to using the @basics alias here.

tariqksoliman (changes requested):

src/essence/Basics/TimeControl_/TimeUI.js
mohatjpl: From what I can tell, when I was resolving merge conflicts these got deleted. I'll be much more careful moving forward about this.

tariqksoliman (approved):
Nice and looks good to me now!

One other minor thing I just found -- and feel free to merge without it -- is that the legend-image.svg (and thereby likely any configured legend from an image) gets rerequested every time any other legend is added or removed from the LegendTool. (You'd need to disable caching to see it rerequest and flash in the legend (with caching on, it's fine and hence why this is a minor issue))

tariqksoliman (approved):
LGTM!

PR #1017, opened 2026-07-15
Add custom GDAL color table support for COG layers via TiTiler

jdrodjpl:
Adds a cogColormapJson layer config field ("Custom Color Table (GDAL)", in its own section of the COG tab) that is passed through to TiTiler's `colormap` query parameter. Accepts either a discrete value-to-RGBA map (e.g. {"1":[10,40,85,255]}) or a list of [[min,max],RGBA] intervals, letting classified rasters (land cover, sea-ice type, etc.) be colored by raw pixel value.

Behavior:
- The color table supersedes the named Colormap (colormap_name), and the Colormap dropdown disables itself with an explanatory tooltip when both a color table and the COG transform are set (via a new generic `disableWhen` hook in the configure app's Maker)
- rescale is suppressed when a color table is set: TiTiler applies rescale before the colormap lookup, which would remap raw values away from the table's keys
- A discrete legend (one labeled color chip per class or interval, suffixed with cogUnits) is generated automatically through the LegendTool/populateCogScale path, with or without the COG transform enabled; fully transparent entries (e.g. nodata) are skipped
- The layer settings panel renders the color table as a swatch strip in place of the named-colormap preview image

### DIFF
--- a/plugins/core/tools/Layers/LayersTool.js
+++ b/plugins/core/tools/Layers/LayersTool.js
@@ -314,6 +314,9 @@ var LayersTool = {
         if (
             layer.type !== 'data' &&
             layer.cogTransform !== true &&
+            // A custom GDAL color table colors raw pixel values without the
+            // COG transform, so it still yields a legend
+            layer.cogColormapJson == null &&
             (layerUrl.startsWith('stac-collection:') ||
                 layerUrl.startsWith('COG:') ||
                 layer.type === 'image')
@@ -336,6 +339,21 @@ var LayersTool = {
             units = layer.variables?.shader?.units ?? ''
         } else {
             units = layer.cogUnits
+        }
+
+        // A custom GDAL color table yields a discrete class legend instead
+        // of a gradient sampled between the rescale min/max (which TiTiler
+        // doesn't apply when a custom color table is set)
+        const colortableLegend = LayersTool.cogColormapJsonToLegend(
+            layer.cogColormapJson,
+            units
+        )
+        if (colortableLegend != null) {
+            L_.layers.data[layer.name]._legend = colortableLegend
+            // The rescale gradation labels don't apply to a color table
+            for (let i = 1; i < 8; i++) $(`[id=tileCogLegend_${i}]`).text('')
+            $('#tileCogColormapMapLines').empty()
+            return
         }
 
         let dynamicLegendConf = []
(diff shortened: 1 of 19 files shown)


plugins/core/tools/Layers/LayersTool.js
```diff
@@ -848,41 +962,21 @@ function interfaceWithMMGIS(fromInit) {
                         supportsExpressions &&
                         node[i].cogExpressionEditable === true
                     ) {
-                        // Check currentCogExpression first (runtime value), then fall back to cogExpression (configured value)
-                        const currentExpression =
```
tariqksoliman, line 852: Why are we deleting the ability for users to set their own band math expressions here?

plugins/core/tools/Layers/LayersTool.js
```diff
@@ -848,41 +962,21 @@ function interfaceWithMMGIS(fromInit) {
                         supportsExpressions &&
```
tariqksoliman, line 962: Looks like a bad merge conflict resolution -- double check whether your new code should be scoped to a`supportsExpressions && node[i].cogExpressionEditable === true` condition.

tariqksoliman (changes requested):

plugins/core/tools/Layers/LayersTool.js
jdrodjpl: This was a merge conflict issue that should be resolved.

plugins/core/tools/Layers/LayersTool.js
jdrodjpl: This was a merge conflict issue that should be resolved.

tariqksoliman (approved):
LGTM! Feel free to merge.

tariqksoliman:
Oh, one thing I forgot: If possible can you add a layer to the Reference-Mission blueprint showing off this feature,

jdrodjpl:
Added to reference mission.

tariqksoliman:
@jdrodjpl I don't see it added to the ref mission blueprint. There should be changes in the following file: https://github.com/NASA-AMMOS/MMGIS/blob/development/blueprints/Missions/Reference-Mission/config.reference-mission.json

In the configure page, be sure to click "Save to Base Blueprint" to actually write your current ref mission config to the shared blueprint one.

PR #1030, opened 2026-08-02
Implement /rename configure endpoint (#962)

### LINKED ISSUE #962: [Bug]: /rename endpoint is unimplemented
The POST /api/configure/rename endpoint is registered but has an empty handler — it does nothing. The configure frontend (configure/src/core/calls.js) has a rename call entry pointing to this endpoint, but the backend at API/Backend/Config/routes/configs.js:881 is just router.post("/rename", function (req, res, next) {});. The endpoint should either be implemented or removed.

From #333

durasi:
## Description
Implements the previously-empty `/api/configure/rename` route (fixes #962), following the existing `/destroy` and `/clone` patterns:
- Validates both `mission` and `newName` with the same name regex and path-traversal guards as `/destroy`
- Rejects renaming to a name that already exists (in the DB or as a `Missions/` directory)
- Updates the `mission` column across all config versions in a single transaction, including the embedded `msv.mission` / `msv.missionFolderName` fields inside each stored config (deep-copied so Sequelize detects the JSON change)
- Renames the `./Missions` directory afterwards, with a partial-success message if the directory move fails (matching `/destroy` behavior)
- Adds the `checkMissionPermission` middleware for consistency with `/upsert` and `/destroy`

Backend-only per the issue; happy to wire up the Configure UI in a follow-up.

## Testing
Verified locally (macOS, Node + Postgres 16, AUTH=none, admin session):
- Happy path: `{"mission":"RenameTest","newName":"RenameTestNew"}` → success; DB shows renamed rows with updated embedded config JSON, unrelated missions untouched
- Path traversal `../evil` → "Invalid mission name."
- Nonexistent mission → "Mission ... not found."
- Existing target name → "A mission named ... already exists."
- Same name → "New mission name must differ from the current name."

### DIFF
--- a/docs/mmgis-openapi.json
+++ b/docs/mmgis-openapi.json
@@ -559,6 +559,66 @@
         }
       }
     },
+    "/api/configure/rename": {
+      "post": {
+        "summary": "Rename a mission configuration",
+        "tags": ["Configure"],
+        "requestBody": {
+          "required": true,
+          "content": {
+            "application/json": {
+              "schema": {
+                "type": "object",
+                "properties": {
+                  "mission": {
+                    "type": "string"
+                  },
+                  "newName": {
+                    "type": "string"
+                  }
+                },
+                "required": ["mission", "newName"]
+              }
+            }
+          }
+        },
+        "responses": {
+          "200": {
+            "description": "Rename mission config response",
+            "content": {
+              "application/json": {
+                "examples": {
+                  "success": {
+                    "summary": "Success",
+                    "value": {
+                      "status": "success",
+                      "message": "Successfully renamed mission to example_mission_renamed"
+                    }
+                  },
+                  "successWithWarnings": {
+                    "summary": "Success with warnings",
+                    "value": {
+                      "status": "success",
+                      "message": "Successfully renamed mission to example_mission_renamed",
+                      "warnings": [
+                        "These missions contain relative paths to example_mission that will need updating: other_mission"
+                      ]
+                    }
+                  },
+                  "failure": {
+                    "summary": "Failure",
+                    "value": {
+                      "status": "failure",
+                      "message": "A mission named example_mission_renamed already exists."
+                    }
+                  }
+                }
+              }
+            }
+          }
+        }
+      }
+    },
     "/api/configure/validate": {
       "post": {
         "summary": "Validate a mission configuration",
--- a/plugins/core/backend/Config/routes/configs.js
+++ b/plugins/core/backend/Config/routes/configs.js
@@ -894,7 +894,130 @@ if (fullAccess)
     }, { full: true, mission: req.body.existingMission });
   });
 
-if (fullAccess) router.post("/rename", function (req, res, next) {});
+if (fullAccess)
+  router.post("/rename", checkMissionPermission, function (req, res, next) {
+    const missionName = req.body.mission;
+    const newName = req.body.newName;
+
+    const nameRegex = /^[A-Za-z0-9_ -]+$/;
+    if (
+      !missionName ||
+      !nameRegex.test(missionName) ||
+      !newName ||
+      !nameRegex.test(newName)
+    ) {
+      logger("error", "Invalid mission name in rename request.", req.originalUrl, req);
+      res.send({ status: "failure", message: "Invalid mission name." });
+      return;
+    }
+    if (missionName === newName) {
+      res.send({
+        status: "failure",
+        message: "New mission name must differ from the current name.",
+      });
+      return;
+    }
+    const missionsBase = path.resolve("./Missions");
+    const resolvedSrc = path.resolve("./Missions/" + missionName);
+    const resolvedDest = path.resolve("./Missions/" + newName);
+    if (
+      (!resolvedSrc.startsWith(missionsBase + path.sep) &&
+        resolvedSrc !== missionsBase) ||
+      (!resolvedDest.startsWith(missionsBase + path.sep) &&
+        resolvedDest !== missionsBase)
+    ) {
+      logger("error", "Path traversal attempt in rename request.", req.originalUrl, req);
+      res.send({ status: "failure", message: "Invalid mission name." });
+      return;
+    }
+
+    Config.findAll({ where: { mission: newName } })
+      .then((existing) => {
+        if ((existing && existing.length > 0) || fs.existsSync(resolvedDest)) {
+          res.send({
+            status: "failure",
+            message: "A mission named " + newName + " already exists.",
+          });
+          return null;
+        }
+        return Config.findAll({ where: { mission: missionName } }).then(
+          (rows) => {
+            if (!rows || rows.length === 0) {
+              res.send({
+                status: "failure",
+                message: "Mission " + missionName + " not found.",
+              });
+              return null;
+            }
+            return sequelize
+              .transaction((t) => {
+                return Promise.all(
+                  rows.map((row) => {
+                    // Deep copy so Sequelize detects the JSON change
+                    const cfg = JSON.parse(JSON.stringify(row.config || {}));
+                    if (cfg.msv) {
+                      cfg.msv.mission = newName;
+                      cfg.msv.missionFolderName = newName;
+                    }
+                    return row.update(
+                      { mission: newName, config: cfg },
+                      { transaction: t }
+                    );
+                  })
+                );
+              })
+              .then(() => {
+                logger(
+                  "info",
+                  "Renamed Mission: " + missionName + " to " + newName,
+                  req.originalUrl,
+                  req
+                );
+                const srcDir = "./Missions/" + missionName;
+                const destDir = "./Missions/" + newName;
+                if (fs.existsSync(srcDir)) {
+                  fs.rename(srcDir, destDir, (err) => {
+                    if (err)
+                      res.send({
+                        status: "success",
+                        message:
+                          "Successfully renamed mission to " +
+                          newName +
+                          " but couldn't rename its Missions directory.",
+                      });
+                    else
+                      res.send({
+                        status: "success",
+                        message: "Successfully renamed mission to " + newName,
+                      });
+                  });
+                } else {
+                  res.send({
+                    status: "success",
+                    message: "Successfully renamed mission to " + newName,
+                  });
+                }
+                return null;
+              });
+          }
+        );
+      })
+      .catch((err) => {
+        logger(
+          "error",
+          "Failed to rename mission: " + missionName,
+          req.originalUrl,
+          req,
+          err
+        );
+        res.send({
+          status: "failure",
+          message: "Failed to rename mission " + missionName + ".",
+        });
+        return null;
+      });
+    return null;
+  });
 
 if (fullAccess)
   router.post("/destroy", checkMissionPermission, function (req, res, next) {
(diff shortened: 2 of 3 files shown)


tariqksoliman:
Hi and awesome!

Yes, a way to rename in the Configure UI would be nice too as well as docs and updating [openapi.json](https://github.com/NASA-AMMOS/MMGIS/blob/development/docs/mmgis-openapi.json).

These are some AI findings that I've truncated down into what I think's relevant:
- Orphans users.missions_managing. Mission-manager (110) permissions are stored as an array of mission names (configs.js:113-127). Rename doesn't rewrite it, so the manager who just renamed the mission immediately loses access to it. Should be updated in the same transaction.
- Force-overwrites msv.missionFolderName. The codebase treats folder name as intentionally separable from mission name (fallback logic at configs.js:190-197, L_.missionFolderName in config.js). If a mission points at a shared/other folder, setting it to newName plus renaming the directory breaks it. Safer: only rename the directory / update missionFolderName when it currently equals the old mission name.
- Cross-mission relative paths break. Clones written via relativizePaths embed ../OldName/... URLs in other missions' configs; those silently 404 after the rename. At minimum warn in the response.
- No tests despite plugins/core/backend/Config/tests/config.spec.js existing, and no Configure UI wiring (author flagged the latter as follow-up). Also minor: fs.existsSync(dest) collision check is TOCTOU, and the DB commit happens before the directory move with no rollback (matches /destroy, so arguably acceptable).

**Also strongly worth considering**: I wonder whether it would be better to have the missionFolderName be completely decoupled from the mission "display name" that users see. It would have to be fully backwards compatible too. I'm not sure which is best either. Thoughts?

durasi:
Thanks for the detailed review — all four points are fair, and the first three are real bugs. Here's my plan for this PR:

1. **users.missions_managing**: update the array inside the same transaction, so a 110 admin keeps access to the mission they just renamed.
2. **missionFolderName**: only rewrite it (and rename the Missions directory) when it currently equals the old mission name; if the mission points at a different/shared folder, leave both untouched.
3. **Cross-mission relative paths**: scan other missions' configs for `../OldName/` references and return them as a warning list in the response, so the operator knows what to fix.
4. **Tests + docs**: add cases to `plugins/core/backend/Config/tests/config.spec.js` and update `docs/mmgis-openapi.json`. I'll also wire up the rename action in the Configure UI — happy to keep that in this PR or split it out, whichever you prefer for review size.

On the TOCTOU check and the commit-before-move ordering: agreed, leaving those as-is to stay consistent with `/destroy`.

On decoupling missionFolderName from the display name: I think that's the right long-term direction — the folder is a storage identifier and the mission name is a label users edit, so tying them means every rename becomes a filesystem migration. A backwards-compatible path could be: keep the existing fallback (missing/empty folder name means "same as mission name") so old configs behave exactly as today, and stop writing the coupling on new operations — then renames become pure metadata changes and the directory never has to move. That's a bigger change than this PR though, and it touches every call site that resolves paths. Fix #2 above is the compatible first step: it stops this endpoint from strengthening the coupling. If you'd like, I can open a separate issue to scope the full decoupling properly.

durasi:
Pushed the fixes. Verified each one locally (Node + Postgres 16, AUTH=none, admin session):

**1. users.missions_managing follows the rename.** Created a 110 user managing `RenameTest2Warned`, renamed the mission, and the array became `{RenameTest2Mgr}`. The update runs inside the same transaction as the config rows.

**2. missionFolderName is no longer force-overwritten.** With the folder name matching the mission name, everything follows as before (mission, msv.mission and folder all became the new name). With the folder set to `SharedFolder`, renaming the mission left the folder as `SharedFolder` and skipped the directory move entirely.

**3. Cross-mission relative paths are reported.** With another mission containing `../RenameTest2Final/somefile.json`, the rename response came back as:
```json
{"status":"success","message":"Successfully renamed mission to RenameTest2Warned","warnings":["These missions contain relative paths to RenameTest2Final that will need updating: OtherMission"]}
```

**4. Tests and docs.** Added four rename cases to `config.spec.js` following the existing pattern (create, act, assert, clean up, skip when not SuperAdmin) and added the endpoint to `docs/mmgis-openapi.json` including the warnings example. I could not run the Playwright suite locally because the test bootstrap requires PostGIS and my local Postgres build does not have it, so those will get their first run in CI.

Configure UI wiring is not in this commit yet. Let me know if you would rather have it here or as a follow-up PR, and I will do it either way.

tariqksoliman:
Awesome! Two things from testing and discussing with AI:

## 1. Concurrent renames into the same target name corrupt the configs table

The collision check (`Config.findAll({where: {mission: newName}})`) runs *outside* the transaction that does the update, so it's a TOCTOU window. Firing these simultaneously:

```
POST /rename {"mission":"DupA","newName":"DupTarget"}
POST /rename {"mission":"DupB","newName":"DupTarget"}
```

both returned `success` (the second with "but couldn't rename its Missions directory"), merging two different missions under one name with a duplicated version number:

```json
{"status":"success","versions":[
  {"mission":"DupTarget","version":0,"createdAt":"...24.193Z"},
  {"mission":"DupTarget","version":0,"createdAt":"...24.202Z"}]}
```

`/configure/get?mission=DupTarget` then resolves to whichever version-0 row Postgres returns first, a later `/upsert` writes v1 on top of the merged history, and `Missions/DupB/` is orphaned on disk. Recovery needs manual SQL.

Fix: do the existence check inside the transaction with a row lock, or add a unique index on `(mission, version)` and treat the violation as the collision case.

## 2. `/rename` accepts names `/add` refuses

`/rename` validates with a whitelist regex, `/add` with a blacklist, so they disagree on what a legal mission name is:

| name | `/add` | `/rename` |
|---|---|---|
| `"   "` (all spaces) | Bad mission name. | **success** |
| `" Padded "` (leading space) | Bad mission name. | **success** |
| `"9lives"` (leading digit) | Bad mission name. | **success** |
| `"Renamé"` / `"Ünicode"` | success | Invalid mission name. |
| name containing a tab or newline | success | Invalid mission name. |
| `"Padded "` (trailing space) | success | success |
| `a+b`, `a%b` | Bad mission name. | Invalid mission name. |

So rename can reach states `/add` blocks on purpose: an all-whitespace mission (stored as `"   "`, creates `Missions/   /`, renders as a blank unclickable entry in the Configure sidebar and landing page), a leading-space name that displays identically to another mission (`" Padded "` renders as `Padded`), and a leading-digit name `/add` rejects via `!isNaN(mission[0])`.

Rename should be at least as strict as `/add` — you shouldn't be able to rename your way into a name you couldn't have created. A `newName !== newName.trim()` check plus non-empty-after-trim covers the gap.

## Separate pre-existing bug this makes more reachable

Three call sites build mission paths from the **display name** instead of `L_.missionPath`:

- `src/essence/Basics/Viewer_/Viewer_.js:187` — `Data/mosaic_parameters.csv`
- `plugins/core/tools/Identifier/IdentifierTool.js:833` — layer data paths
- `plugins/core/tools/Measure/MeasureTool.js:1468` — DEM paths

These are broken **today**, with no rename involved, for any mission whose `missionFolderName` differs from its name — which is a supported setting, editable right on the Configure Home tab under "Paths → /Missions Folder Name". #772 converted `Layers_.js`, `LandingPage.js` and `essence.js` when it introduced `missionFolderName` and missed these three; 31 other call sites use `L_.missionPath` correctly, and `Viewer_.js` itself uses it at lines 308 and 331. Worth a small separate PR — it's a live bug, and this PR makes folder divergence more likely to be exercised.

## For the Configure UI follow-up

`configure/src/core/calls.js` already has a `rename` entry, so this is mostly a `RenameConfigModal` copied from `CloneConfigModal` plus a fifth icon in the mission header toolbar. Two things worth having beyond a text input:

- An **"also rename the `/Missions` folder" checkbox** — `followFolder` chosen explicitly by the operator rather than inferred, defaulted on when the folder currently matches the name and disabled with an explanation when it doesn't.
- Somewhere to **display the returned `warnings`**, otherwise the cross-mission path warning never reaches the person who needs to act on it.

It should also warn that renaming the folder breaks data loading for anyone with the mission open until they reload (there's no websocket broadcast on rename, so open clients don't recover on their own), and that renaming the mission at all breaks existing `?mission=OldName` links and bookmarks permanently. The delete modal's "None of the mission's data files in /Missions will be deleted" is a good precedent for that kind of note.

durasi:
Thanks, and good catches. Both are fixed and rebased onto current development.

## 1. Concurrent renames

Your reproduction was exactly right, and I could reproduce it here. One thing though: I do not think a row lock closes this window. The check is for the *absence* of rows, and `SELECT ... FOR UPDATE` only locks rows that exist, so two callers would still both see an empty result and both proceed. It is a phantom read rather than a row conflict.

So I took a transaction scoped advisory lock keyed on the target name, then re-check inside the transaction:

```js
await sequelize.query("SELECT pg_advisory_xact_lock($1, $2)", { bind: lockKeys, transaction: t });
```

The keys come from a sha1 of `mission_rename:<newName>`, so renames into the same name serialize on the same lock and renames into different names do not block each other. The lock is released when the transaction ends, so there is nothing to clean up. The outer check stays as a fast path.

Verified locally, three simultaneous renames into one target:

```
{"status":"failure","message":"A mission named RaceTarget already exists."}
{"status":"success","message":"Successfully renamed mission to RaceTarget"}
{"status":"failure","message":"A mission named RaceTarget already exists."}
```

The database afterwards shows one winner and the losers untouched, no merged history:

```
  mission  | version | msv_mission | folder
-----------+---------+-------------+----------
 DupB      |       0 | DupB        | DupB
 DupTarget |       0 | DupTarget   | DupTarget
```

I also confirmed the lock actually blocks rather than the test passing on timing: a second session waiting on the same key took 2683 ms behind a 3 second holder.

On the unique index option: that would also work and is arguably the stronger guarantee, but it changes the shape of the table for `/upsert` and `/clone` as well, and the model has `unique: false` on both columns today. Happy to open a separate issue for it if you want that as a follow up.

## 2. Name validation

Rename now applies the same rules as `/add`, so the three cases in your table are rejected:

```
[   ]        -> {"status":"failure","message":"Bad mission name."}
[ Padded ]   -> {"status":"failure","message":"Bad mission name."}
[9lives]     -> {"status":"failure","message":"Bad mission name."}
```

One note on scope: your suggested `newName !== newName.trim()` check also rejects a trailing space, which `/add` accepts today. I kept it, since the same argument applies and two names that render identically are a hazard either way, but say the word if you would rather it match `/add` exactly.

The check applies to `newName` only. A mission that already has a bad name should still be renamable out of it.

## Tests

Added two cases to `config.spec.js`: the strict name rejections, and a concurrency test that fires two renames at one target with `Promise.all` and asserts exactly one succeeds and the loser survives under its old name. Same caveat as last time, the Playwright suite needs PostGIS which my local Postgres build does not have, so these get their first run in CI.

## The other two items

The three call sites building paths from the display name look like a real bug, and separate from this endpoint, so I would rather not fold it in here. Happy to take it as its own PR if nobody else is on it.

For the Configure UI, your outline is clear enough to build from, including the explicit followFolder checkbox and surfacing the warnings. Do you want it as a follow up PR, or in this one?

tariqksoliman (approved):
Thank you!

And both the "Configure UI" and "three mission path call site bugs" are good candidates for follow-up PRs.

tariqksoliman:
Don't worry about the failing checks; they're workflow permission issues that I have to fix eventually.

durasi:
Thank you so much I'm creating a new branch and starting to work on Configure UI and three mission path call site bugs right now. I'm doing a thorough review of the details and will open the follow up PRs soon

PR #1039, opened 2026-08-14
Use L_.missionPath in three mission data path call sites (#1038)

### LINKED ISSUE #1038: [Bug]: Three call sites build mission paths from the display name instead of L_.missionPath
### Checked for duplicates

Yes - I've already checked

### Describe the bug

Three call sites build mission data paths from the mission display name instead of `L_.missionPath`:

- `src/essence/Basics/Viewer_/Viewer_.js:187` (`Data/mosaic_parameters.csv` lookup path)
- `plugins/core/tools/Identifier/IdentifierTool.js:833` (layer data paths)
- `plugins/core/tools/Measure/MeasureTool.js:1468` (DEM paths)

All three do `'Missions/' + L_.mission + '/' + url`, while `L_.missionPath` is set to `'Missions/' + L_.missionFolderName + '/'` in `src/essence/Basics/Layers_/lifecycle/config.js:50`.

For any mission whose `missionFolderName` differs from its name the three paths resolve to a directory that does not exist, so the mosaic parameters lookup, identifier data queries and measure profiles all fail. Divergence is a supported configuration, editable on the Configure Home tab under Paths, /Missions Folder Name, and #771 deliberately allowed the folder name to drift from the mission name.

#772 introduced `missionFolderName` and converted `Layers_.js`, `LandingPage.js` and `essence.js`, but missed these three. 30 other call sites use `L_.missionPath` correctly, and `Viewer_.js` itself uses it at lines 308 and 331, so the fix is to bring these three in line.

No rename is needed to hit this. It was raised during review of #1030 (https://github.com/NASA-AMMOS/MMGIS/pull/1030#issuecomment-5283921984), which makes folder divergence easier to reach but is not the cause.

### What did you expect?

The three call sites should use `L_.missionPath`, which already carries the trailing slash, so paths resolve against the configured mission folder rather than the display name.

### Reproducible steps

1. Create a mission and set Paths, /Missions Folder Name to a value different from the mission name, with the data living in the folder.
2. Open a feature with imagery to trigger the `Viewer_` mosaic parameters lookup, run an Identifier query against a layer with a relative data URL, or draw a Measure profile over a DEM with a relative path.
3. The request goes to `Missions/<displayName>/...` rather than `Missions/<folderName>/...` and 404s.

### Environment

Present on `development` at 24fae496.

durasi:
Fixes #1038.

Three call sites built mission data paths from the mission display name rather than the configured mission folder:

- `Viewer_.js` mosaic parameters lookup path
- `IdentifierTool.js` layer data paths
- `MeasureTool.js` DEM paths

All three now use `L_.missionPath`, which is set to `'Missions/' + L_.missionFolderName + '/'` in `Layers_/lifecycle/config.js` and already carries the trailing slash, so the concatenations lose their separate `'/'`.

This matches the 30 other call sites that already resolve through `L_.missionPath`, including two in `Viewer_.js` itself. #772 introduced `missionFolderName` and converted `Layers_.js`, `LandingPage.js` and `essence.js`, but these three were missed.

Raised during review of #1030 and split out as a follow up, as suggested there.

## Testing

Change is a substitution of an equivalent expression: for a mission whose folder name matches its name, `L_.missionPath` produces the same string the old concatenation did, so existing behaviour is unchanged. For a mission whose folder name differs, the paths now resolve against the folder instead of the display name, which is the fix.

I did not exercise the three code paths end to end locally, since each needs mission data in place (a mosaic parameters CSV, a layer with a relative data URL, a DEM). Happy to set that up if you would like it verified that way before merging.

### DIFF
--- a/plugins/core/tools/Identifier/IdentifierTool.js
+++ b/plugins/core/tools/Identifier/IdentifierTool.js
@@ -830,7 +830,7 @@ function queryDataValue(url, lng, lat, numBands, layerUUID, callback) {
     } else if (url.startsWith('/vsicurl/') || url.startsWith('Missions/')) {
         dataPath = url
     } else {
-        dataPath = 'Missions/' + L_.mission + '/' + url
+        dataPath = L_.missionPath + url
     }
 
     dataPath = IdentifierTool.fillURLParameters(dataPath, layerUUID)
--- a/plugins/core/tools/Measure/MeasureTool.js
+++ b/plugins/core/tools/Measure/MeasureTool.js
@@ -1465,7 +1465,7 @@ function makeProfile() {
         // enable remote access via GDAL Virtual File Systems /vsi* prefix
         let pathDEM
         if (path.startsWith('/vsi')) pathDEM = path
-        else pathDEM = 'Missions/' + L_.mission + '/' + path
+        else pathDEM = L_.missionPath + path
 
         //elevPoints.push([{"x": clickedLatLngs[numOfPts - 2].x, "y": clickedLatLngs[numOfPts - 2].y}, {"x": clickedLatLngs[numOfPts - 1].x, "y": clickedLatLngs[numOfPts - 1].y}]);
         elevPoints = [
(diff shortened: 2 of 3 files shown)


tariqksoliman (approved):
LGTM and thanks!

PR #1041, opened 2026-08-14
Add rename action to the Configure UI (#1040)

### LINKED ISSUE #1040: [Bug]: Configure UI has no rename action although the endpoint and calls.js entry exist
### Checked for duplicates

Yes - I've already checked

### Describe the bug

`POST /api/configure/rename` was implemented in #1030, and `configure/src/core/calls.js` has had a `rename` entry pointing at it for some time, but nothing in the Configure UI calls it. The mission toolbar on the Home tab offers export, upload, clone and delete, so renaming a mission still means going through the API directly.

### What did you expect?

A rename action in the mission toolbar, following the existing `CloneConfigModal` and `DeleteConfigModal` patterns. Beyond a text input, two things are worth having, both raised in the review of #1030 (https://github.com/NASA-AMMOS/MMGIS/pull/1030#pullrequestreview-4931871615):

- An "also rename the /Missions folder" checkbox, so `followFolder` is chosen explicitly by the operator rather than inferred. Default it on when the folder name currently matches the mission name, and disable it with an explanation when it does not, since that folder may be shared with another mission.
- Somewhere to show the warnings the endpoint returns. It reports other missions holding relative paths into the renamed one, and right now that never reaches the person who has to act on it.

The modal should also note that renaming the folder stops data from loading for anyone with the mission open until they reload, since there is no websocket broadcast on rename, and that renaming the mission breaks existing `?mission=OldName` links and bookmarks permanently. The delete modal's note that data files in /Missions are not deleted is a good precedent for that kind of warning.

Honouring the checkbox needs a small endpoint change too: `/rename` currently infers `followFolder` and takes no such field.

### Reproducible steps

1. Open the Configure UI on any mission and look at the toolbar on the Home tab.
2. There is no rename action, although `calls.js` defines one and the endpoint exists.

### Environment

Present on `development` at 24fae496.

durasi:
Fixes #1040.

Adds a rename action to the mission toolbar on the Home tab, following the existing clone and delete modals.

## Folder decision is explicit

The checkbox is on by default when the mission's `/Missions Folder Name` matches the mission name, and disabled with an explanation when it does not, since that folder may be shared. The endpoint takes a new optional `followFolder` field so the operator's choice is honoured, within the same safety rule as before: the folder is only moved when it currently matches the mission name. Omitting the field keeps the previous behaviour, so existing callers are unaffected.

## Warnings are surfaced

`/rename` reports other missions holding relative paths into the renamed one. Those are now shown in the snackbar rather than dropped.

## Warnings in the modal

The modal states that renaming the folder stops data from loading for anyone with the mission open until they reload, that renaming breaks existing `?mission=OldName` links permanently, and that data files in /Missions are not deleted.

## Testing

Verified locally against the built Configure UI (Node, Postgres 16, AUTH=none, admin session):

- Rename icon appears between clone and delete, opens the modal.
- With the folder name matching the mission name, the checkbox is on and editable.
- After setting `/Missions Folder Name` to `SharedFolder` and saving, reopening the modal shows the checkbox disabled with the explanation naming the current folder.
- Renaming in that state left the folder alone. Both config versions moved to the new name in `mission` and in the embedded `msv.mission`, while `missionFolderName` stayed as it was:

```
    mission    |      msv      |    folder
---------------+---------------+--------------
 UITestRenamed | UITestRenamed | UITest
 UITestRenamed | UITestRenamed | SharedFolder
```

The v0 row keeps `UITest` because that is what the folder was when that version was saved.

Client-side guards reject an empty name, a name equal to the current one, and a name already in the mission list, so those never reach the endpoint. Server-side validation from #1030 still applies to everything else.

### DIFF
--- a/configure/src/components/Tabs/Home/Home.js
+++ b/configure/src/components/Tabs/Home/Home.js
@@ -18,11 +18,13 @@ import BrowserUpdatedIcon from "@mui/icons-material/BrowserUpdated";
 import ContentCopyIcon from "@mui/icons-material/ContentCopy";
 import UploadIcon from "@mui/icons-material/Upload";
 import DeleteForeverIcon from "@mui/icons-material/DeleteForever";
+import DriveFileRenameOutlineIcon from "@mui/icons-material/DriveFileRenameOutline";
 import SaveIcon from "@mui/icons-material/Save";
 
 import config from "../../../metaconfigs/tab-home-config.json";
 import UploadConfigModal from "./Modals/UploadConfigModal/UploadConfigModal";
 import CloneConfigModal from "./Modals/CloneConfigModal/CloneConfigModal";
+import RenameConfigModal from "./Modals/RenameConfigModal/RenameConfigModal";
 import DeleteConfigModal from "./Modals/DeleteConfigModal/DeleteConfigModal";
 
 const REFERENCE_MISSION_NAMES = new Set([
@@ -154,6 +156,13 @@ export default function Home() {
       }),
     );
   };
+  const handleRename = () => {
+    dispatch(
+      setModal({
+        name: "renameConfig",
+      }),
+    );
+  };
   const handleDelete = () => {
     dispatch(
       setModal({
@@ -225,6 +234,16 @@ export default function Home() {
                 <ContentCopyIcon fontSize="medium" />
               </IconButton>
             </Tooltip>
+            <Tooltip title={"Rename Mission"} placement="bottom" arrow>
+              <IconButton
+                className={c.renameIcon}
+                title="Rename"
+                aria-label="rename"
+                onClick={handleRename}
+              >
+                <DriveFileRenameOutlineIcon fontSize="medium" />
+              </IconButton>
+            </Tooltip>
             <Tooltip title={"Delete Mission"} placement="bottom" arrow>
               <IconButton
                 className={c.deleteIcon}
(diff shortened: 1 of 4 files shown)


tariqksoliman:
Nice! Two things though:
1. Once I rename a mission, in the UI I properly see the mission list on the left get updated -- but the currently active mission configuration retains the old name. It should refresh the currently active mission in the configure page after a rename.

2. Could you make
> Renaming the folder stops data from loading for anyone with this mission open until they reload. Renaming the mission also breaks existing ?mission=UIKEEPRENAMED2 links and bookmarks permanently. None of the mission's data files in /Missions will be deleted.
visually look much more like a warning, maybe with a ⚠️at least. There's also the case where a mission does not have `msv.missionFolderName` set which would make a rename temporarily detach a mission from its data.

Thanks!

durasi:
Thanks Tariq, both addressed.

1. Active config not refreshing after rename. The rename success handler now dispatches setMission(newName) after the missions list is refetched. Since Main's mission effect refetches the configuration whenever the active mission changes, the open Configure page reloads under the new name on its own, no page refresh. Verified locally: after renaming, the Home title, the /Missions Folder Name field, and the left mission list all move to the new name in place.

2. Warning styling and the unset folder case. The warning is now an actual warning block with a ⚠️ icon, an amber border, and a highlighted line for the permanent ?mission= link breakage. The wording also adapts: when the folder does not follow the rename (folder name differs, or the operator unchecks the box), it states the /Missions folder is left as it is and the data stays where it currently sits, which covers the case where a mission has no msv.missionFolderName set and a rename would otherwise detach it from its data.

Rebased onto development. Let me know if you'd like the wording tightened further.

tariqksoliman (approved):
LGTM!

PR #1045, opened 2026-08-19
feat(auth): add typed API authentication with explicit public mounts

midatm1234:
Machine-consumed endpoints currently receive login HTML or HTTP-200 failure envelopes when authentication fails. This adds `ensureUserForApi(options)` with typed HTTP 401/503 responses and adopts it in the core Shortener mount.

The gate requires credentials by default, including on `AUTH=off` and `AUTH=none` deployments. A mount must explicitly pass `{ allowPublic: true }` to allow anonymous access in those two modes. Shortener opts in to preserve public link sharing. Local/CSSO deployments still require an authenticated identity or a valid long-term token.

The implementation shares the session-permission predicate with legacy `ensureUser()`, reuses the existing token lookup/hydration path, handles token-query rejection, and removes token material from legacy rejection logs. Typed errors have configurable codes/messages, and successful token/public-session requests receive stable hashed identities for request correlation. Legacy middleware signatures and response contracts remain intact; Shortener deliberately adopts the new error contract.

The final diff is authentication-only: middleware, one core caller, documentation/specs, and tests. It adds no frontend API or action registry. Current `development` (`bfadcd8`) is included, resolving the previous merge conflict.

**Validation — 2026-09-09, `7c05d01`**

- Auth and real Shortener-mount tests: **16 passed**, without a database, built frontend, or Reference Mission.
- Combined host and companion plugin unit run: **292 passed**, zero skipped, unexpected, or flaky results.
- Companion raster integration: **2 passed**, with the manifest's exact Python dependencies.
- Integrated production build: passed on Node 24.18.0.
- Focused auth lint, Node syntax checks, and `git diff --check development...HEAD`: passed.

Standalone host verification:

```bash
npx cross-env PLAYWRIGHT_TEST_UNIT_ONLY=true playwright test tests/unit/apiAuthentication.spec.js tests/unit/shortenerApiAuth.spec.js --project=chromium --workers=1
```

The tests exercise isolated Express servers and the real Shortener mount with a terminal router in place of database writes. A live authenticated deployment and external identity provider were not exercised locally.

[MMGIS-Plugins#6](https://github.com/NASA-AMMOS/MMGIS-Plugins/pull/6) consumes this typed-authentication contract and should merge after this PR. Host verification is self-contained and does not require that plugin PR.


tariqksoliman:
Still a draft, I know but I iterated on a review for it:

## Verdict: which parts are limitations, which are anti-patterns

Genuine host limitations, all small and worth fixing on their own:

1. **HTTP-200 failure envelopes.** `stopGuests` and most core handlers answer failures with a 200,
   which is documented in `plugins/core/backend/README.md` as a thing that "surprises people". A
   machine-consumed endpoint genuinely cannot express 401/503. The typed gate is a defensible host
   addition — after fixing the `AUTH=off` divergence flagged in the first review.
2. **`validateLongTermToken` had no error path** — a rejected query left the request unanswered.
   Real bug, fixed here.

That is the entire list. Anti-patterns:

3. **The registry is mounted on the external embedding API** rather than a plugin-facing module,
   which both misdirects plugin authors and exposes every registered plugin action to the embedding
   page.
4. **A parallel imperative registry** where the repo's documented answer is a declarative manifest
   family plus static validation. This is the core objection.
5. **Host absorbs a plugin-specific concept.** "Copilot" is one plugin's vocabulary. The host now
   ships `registerCopilotAction`, a bespoke schema dialect, and 512 KiB/128-action budgets for a
   single out-of-repo consumer, with zero in-repo callers of either new contract.
6. **Hardening framed as a trust boundary that does not exist.** Deep-freezing, error redaction,
   node/byte/depth caps — against build-time-linked plugins that already own `window`. #6's own
   limitations admit "Installed frontend plugins remain trusted host code". Useful as an LLM I/O
   contract; not security, and it reads as security.
7. **Cross-repo SHA pinning.** #6's CI checks out MMGIS and overlays two exact #1045 commits. The
   host PR is therefore unreviewable in isolation and the plugin CI silently rots the moment #1045
   is rebased or amended in review.
8. **The split leaves the bottom slice unverifiable.** Splitting 25k lines was right, but the host
   slice was cut so that all its consumers are outside the repo; its only in-repo test of the
   browser facade self-skips without `build/index.pug` and a Reference-Mission.

## How I'd implement it

Two reframings drive everything below. First, the untrusted party is the **model**, not the plugins:
plugins are build-time-linked host code that already own `window`, so the validation, bounding, and
allowlisting belong at the boundary where model output enters the app — inside the Agent plugin.
Second, what Agent actually needs is to see which tools are present and usable and then call them,
and it can already do that from inside itself. `ToolController_` knows the mission's configured tool
list, holds the loaded tool modules, and exposes `getTool`, `openTool`, and `closeTool`; each tool
module carries its own state and public methods; `L_` says which layers exist and are on. That is
discovery and invocation, already available to in-bundle code with no host change.

Crucially the dependency runs one way only: **Agent depends on the tools it wants to drive, and no
tool depends on Agent.** AgentTool and AnalysisTool each work standalone. That is the direction that
makes MMGIS-Plugins#7 unnecessary rather than merely smaller.

**Land the host auth work as its own PR, with no Copilot vocabulary in it.** A typed gate for
machine-consumed endpoints plus the `validateLongTermToken` error path plus the token no longer
being written to the log. Two changes to what #1045 has today: authentication should be an explicit
per-mount decision rather than something that silently opens up when the deployment sets
`AUTH=off`, and `ensureUser` should be refactored onto the same session predicate so the
`{111,110,001}` permission set exists once. This PR has in-repo callers the moment any core route
adopts it, stands on its own merit, and is reviewable without reference to another repository.

**Keep the whole catalog inside the Agent plugin, as adapters it owns.** Agent holds one module: a
table of the tools it knows how to drive, each entry pairing a model-facing action description with a
handler and a predicate for whether the action is usable right now. The predicate asks
`ToolController_` whether the tool is configured and loaded and asks the tool module or `L_` whether
its preconditions hold; if the answer is no, the action simply isn't offered to the model that turn.
The handler calls the tool's public methods and the host singletons — `L_`, `F_`, `ToolController_` —
the same way every other plugin does. So `open_animation_tool` becomes a one-line handler instead of
something deferred pending a host facade, and Analysis needs no patch at all: Agent detects it and
calls it. Zero host change, one repo, and a mission without AgentTool installed is bit-for-bit
unaffected.

The cost of this shape is that Agent needs the tools' public methods to be a reasonably stable
contract, and that adding a tool to the catalog is an Agent-side edit. Both are fine at this scale,
and both are visible and testable in one place, which the runtime registry is not.

**Only promote this to a host concern for tools Agent cannot know about.** The one thing adapters
cannot do is let a third-party plugin the Agent has never heard of advertise itself. That is a real
future need and the only honest justification for host involvement — and when it arrives, the answer
is the repo's existing declarative mechanism, not a runtime lookup: do it the way the existing
families work. An action is declared in `plugin.json` with
its id, category, applicability metadata and its argument schema; the build step discovers enabled
plugins, enforces `pluginDependencies`, and generates the registry file with static imports; a small
runtime dispatcher reads that generated data with no hardcoded ids; and `plugin-cli validate`
enforces the schema dialect, id collisions, and staleness statically without loading anything. Two
consequences worth noting: nearly all the bounding in #1045 becomes a build-time check (descriptors
are static, so there is no need to deep-freeze them or budget their serialized size at runtime), and
the family should be named for what it is — actions for an assistant — not for one vendor's product.

**Validate model output where it arrives, with an off-the-shelf validator.** The arguments a model
produces are the untrusted input, and they arrive in the Agent backend. Ajv is already resolvable in
this tree, so the bespoke JSON Schema subset can go away along with the silent-rejection surprises
it creates for plugin authors (`$ref`, `pattern`, `format`, composition keywords). Keep exactly two
runtime protections on the frontend, because they are about model and handler behavior rather than
trust: a handler timeout with an abort signal, and a cap on result size before it goes back into a
prompt. Drop the redaction and deep-freeze framing, or document it plainly as "don't let a buggy
handler corrupt the catalog", which is what it actually does.

**Nothing new on `window.mmgisAPI` beyond what an embedder wants.** `setLayerOpacity` is a
reasonable addition to the embedding API on its own merit and can go in whenever; the four
`*CopilotAction` methods should not be there at all.

**Sequence it so each PR is verifiable where it lives.** Host auth first, self-contained, with an
in-repo caller. Then Agent, AgentChat, and the adapter catalog in MMGIS-Plugins against
`development` HEAD rather than an overlay of pinned host commits, so plugin CI stops depending on an
unmerged branch's SHAs. No Analysis PR. If the manifest family is ever wanted in core later, it
lands with a core example action so the host has an in-repo consumer, and with the browser test
wired to a CI mode that actually builds and runs it rather than skipping.

PR #1047, opened 2026-08-20
Add worksOnMobile flag for mobile enabled tools

ac-61:
## Purpose
- This makes it so the tools that have a worksOnMobile field can be enabled in mobile mode.

## Proposed Changes
- [ADD] Add worksOnMobile flag for mobile enabled tools


tariqksoliman (approved):
LGTM too, thanks!

PR #1057, opened 2026-08-25
Added leaflet wraparound for lat/lon display

grallewellyn:
## Purpose
Confusing for users to have a longitude value greater than 180 or less than -180. The purpose of this is to provide wraparound for the lat/lon visual display in the bottom right corner 
Screenshot 2026-08-25 at 10 20 30 AM

## Proposed Changes
- [CHANGE] Longitude wraps around in display using built in leaflet function 
## Issues
Discussed on slack 
## Testing
Tested locally going to the far left of the screen and the far right with and without "World Copy Jump" turned on (this is configured in the config -> User Interface page)
Confirmed that the longitude never goes over 180 or under -180

### DIFF
--- a/src/essence/Basics/UserInterface_/components/Coordinates/Coordinates.js
+++ b/src/essence/Basics/UserInterface_/components/Coordinates/Coordinates.js
@@ -702,8 +702,10 @@ function changeCoordType(newCoordIndex) {
 
 //Update mouse lat long coordinate display on mouse move
 function mouseLngLatMove(e) {
-    Coordinates.mouseLngLatRaw = [e.latlng.lng, e.latlng.lat]
-    Coordinates.mouseLngLat = [e.latlng.lng, e.latlng.lat]
+    // Normalize coordinates to -180/180 range using Leaflet's built-in method
+    const normalized = Map_.map.wrapLatLng(e.latlng)
+    Coordinates.mouseLngLatRaw = [normalized.lng, normalized.lat]
+    Coordinates.mouseLngLat = [normalized.lng, normalized.lat]
     if (
         L_.configData.coordinates &&
         L_.configData.coordinates.coordelev === true


lestarch-autobot (commented):
### Runtime verification — Earth fixed, Lunar South Pole unaffected

Tested end-to-end on a local dev server (`:8889`) by hovering the bottom-right readout, comparing `development` (pre-fix) against this PR head `b4461625` at the **identical map view, zoom and hover pixel**.

**Earth (`Reference-Mission`, EPSG:3857) — the fix works**

Zoomed out until multiple world copies were visible, then hovered the 3rd copy east at the same pixel on both branches:

| Before (`development`) | After (this PR) |
|---|---|
| `932.34375000°, 8.40716816°` | `-147.65625000°, 8.40716816°` |

`-147.65625 = 932.34375 − 1080`, i.e. exactly 3 wraps. Console: `map.options.crs.wrapLng === [-180,180]`, `map.wrapLatLng({lat:0,lng:200}).lng === -160`.

**Lunar South Pole (custom `IAU2000:30120`) — inert, and it doesn't matter**

`L.Proj.CRS` (`src/external/Leaflet/proj4leaflet.js`) only does `includes: L.CRS` and never sets `wrapLng`, so `wrapLatLng()` is a clone-only no-op for any mission with `projection.custom === true`. Confirmed live on the loaded mission:

```json
{"crs":"IAU2000:30120","mapWrapLng":"undefined","globalWrapLng":"undefined","wrapTest":200}
```

(`wrapTest` = `map.wrapLatLng({lat:-85,lng:200}).lng` → unchanged.)

Tried hard to produce |lon| > 180: min zoom, repeated ~700px drags east and west past the antimeridian, hovers at all canvas corners and near the pole. Real-hover extremes on this PR: **-179.93 … 179.88** (127 samples). A 19,881-point sweep of `containerPointToLatLng` (±3000px around the home view — the same conversion the mousemove handler consumes) gives **min -179.5166 / max 179.6615**, *identical on both branches*. The polar-stereographic inverse already returns wrapped longitudes, so this mission never showed the bug and the PR changes nothing for it.

**Regression checks**

- Lunar: live coordinate updates OK; elevation readout OK (`545.000m` from `LRO_LOLA_DEM_4000m.tif`) with the normalized value now feeding the `getbands` query.
- Earth: elevation OK (`94.776m` at `-122.46459961°, 37.76202989°`); note the demo DEM footprint is small (lon -122.507..-122.414, lat 37.710..37.781).
- Not covered: whether the fix also repairs elevation lookups inside wrapped world copies (likely a bonus); the Cesium/Globe readout path (outside this diff).

**Optional nits, none blocking**

1. `mouseLngLatRaw` is no longer raw and it feeds the elevation query — worth renaming, or wrapping only `mouseLngLat`.
2. Two other display paths still show unwrapped longitude: the mobile map-center text in `Coordinates.js` and `L_.Coordinates.setCoords` from `Globe_.js`.
3. `tests/e2e/map/coordinates.spec.js` could gain a cheap `|lon| <= 180` assertion after panning past the antimeridian.
4. Nice side effect: wrapping happens before `coordlngoffset` is applied, so a mission using offset 180 now stays within 0–360 instead of drifting past 360.

Verdict: **helps on Earth, provably inert on the Lunar South Pole mission, no regressions.**

[Written by Devin](https://nasa-jpl-demo.devinenterprise.com/sessions/20497bbcb3ed4a5dbdadb5ca1b958195)

tariqksoliman (approved):
LGTM and thanks!

PR #1076, opened 2026-09-09
Fix for geodatasets filter handling

tariqksoliman:
Hardens value handling in the geodatasets in filter operator on /api/geodatasets/get so every user-supplied value is bound strictly as data, and bounds the size of the value list.

[FIX] in filter replacement placeholders are named only from numeric indexes `(filter_value_<i>_<j>);`  

[CHANGE] in lists are capped at 500 values / 1000 chars per value `(MAX_IN_VALUES, MAX_IN_VALUE_LENGTH);` excess is truncated.


tariqksoliman:
Superseded by #1084

PR #1087, opened 2026-09-15
#1070 - Highlight resolved values in the Legend Tool on Identifier hover

### LINKED ISSUE #1070: [New Feature] - Identifier Tool - Mousing over should also highlight values in the Legend Tool.
As a user, when I mouse over the map with the Identifier Tool active, I want to see the layer's respective legends in the Legend Tool also highlighted.

Do note that while plugins cannot depend on each other, it is still in the patten that they can check for the presence of another plugin and behave and call the other accordingly -- they just must not assume the other plugin is always there or enabled.

Raydman:
## Purpose
When the Identifier tool is active, mousing over the map resolves a value for
each raster layer under the cursor. The Legend tool shows what those values
mean, but nothing connected the two — you had to match the reading to its
legend entry by eye. This lights up the entry the cursor just resolved.

## Proposed Changes
- [ADD] `plugins/core/tools/Legend/legendHighlight.js` — which legend entry a
  value belongs to, as a module importing nothing from `src/essence` so it can
  be unit tested
- [ADD] `LegendTool.highlightEntries(matches)` and `clearHighlightedEntries()`,
  taking `[{ layerUUID, value }]`
- [ADD] Legend entries carry `data-legend-layer-uuid`, `data-legend-value` and
  `data-legend-property-value`, so an entry can be addressed; continuous scales
  record their values in paint order and mark the matched band on the ramp
- [CHANGE] `IdentifierTool` reports what each layer resolved to, from both the
  colour-match path and the asynchronous true-value read
- [CHANGE] `queryDataValue` reports back exactly once on every route out
- [ADD] `tests/unit/legendHighlight.spec.js` and
  `tests/e2e/tools/legend-highlight.spec.js`

## Issues
- Closes #1070

## Design notes

**Why the tools are not dependencies of one another.** The Legend is reached
through `ToolController_` with an optional call, and is deliberately not added
to `pluginDependencies` — per `plugins/core/interactions/README.md` an
unresolved dependency drops the Identifier from the registry entirely. A
mission may enable either tool alone, so the call is a no-op when the other
tool is absent.

**Why matches are kept per layer.** The true-value reads land independently,
so each layer's result is stored as it arrives and the legend is only
republished once none are outstanding. Publishing an empty set while a read is
in flight is what made the marker blink.

**Why a read reports back even when it fails.** Callers count the reads they
are waiting on. A route that returned silently left the count above zero
forever, so the legend was never republished and kept the previous marker while
the readout had already blanked. Every route out of `queryDataValue` now
reports a value or `null`, exactly once — the once-only guard also closes a
pre-existing path where a string payload could invoke the callback twice.

**Why each hover carries a generation.** `activeLayerNames` is rebuilt on every
hover and the matches are shared, so a slow response could write into a later
hover's set, key a value to the wrong layer, or re-light a legend that
`clearCursor` had just cleared. Each read now captures the layer it was issued
for and the hover it belongs to, and drops itself if that hover is over.

**Why nearest-match applies only to ramps.** A ramp's labels are sampled points
along a continuum, so a reading between two of them still belongs to one.
Discrete rows are unrelated values that happen to be numbers, and a legend may
hide some of them, so nearest would light whichever row a reading happened to
sit closest to. Rows therefore require an exact match — with one addition: a
label naming a span (`10 - 20`) matches a reading inside that span, which is
containment rather than snapping, and is what makes a binned legend usable.
Happy to drop the span handling if you would rather rows were exact and nothing
else.

**Why the marker uses `mix-blend-mode: difference`.** A tinted bar disappears
where the ramp is its own colour. A white bar inverted against whatever is
beneath it stays visible everywhere, and it matches what the Legend already
does for its scale labels. The transition is gone, so the marker lands where
the cursor is rather than easing after it.

**Why `propertyValue` is recorded alongside the label.** A legend can label an
entry with text while the map is styled by a number. The existing ramp tooltip
already prefers `propertyValue`, so a match has to be able to see it too.

**Why nodata is unwrapped.** Readings arrive array-wrapped, and a masked pixel
arrives as `[null]`, which multiplies out to `0` rather than `NaN` — nodata was
being marked as a real reading at the bottom of the ramp.

## Testing

`tests/unit/legendHighlight.spec.js` — 21 cases over span parsing, exact and
nearest resolution, `propertyValue`, and candidate fallback. It pins the
behaviour you flagged: rows `1`, `2`, `50` with a reading of `27` light
nothing.

`tests/e2e/tools/legend-highlight.spec.js` — 9 cases. Builds a known legend
through the public `overwriteLegends` API rather than relying on mission data,
then checks hover → row lit / ramp marked → cleared, that a redrawn legend
keeps its highlight, and that closing the Identifier and moving the cursor off
the map both clear it. These fail against the previous build, which is what
showed the nearest-match and marker changes had landed.

Band resolution checked against live readings from the USGS SF DEM through
titiler-pgstac, on macOS (arm64) / Chromium:

| Reading | Band | Marker | Nearest stop |
|---|---|---|---|
| 233.05 m | 3 | 38.89% | 250 m |
| 80.91 m | 6 | 72.22% | 100 m |
| 31.81 m | 7 | 83.33% | 50 m |
| 88.70 m | 6 | 72.22% | 100 m |

Also verified: production build compiles clean, and ESLint reports no new
errors — the two `mmgisglobal` `no-undef` errors in `IdentifierTool.js` are
present on `development` already.

### Correction on the earlier test-suite note
My first description claimed the Legend and Identifier specs were broken on
`development` because `tests/global-setup.js` omits the mission name. That was
wrong, and you were right: `createReferenceMission` derives the name from the
variant, and `global-setup` does create an admin, log in, and send the session
cookie. The 404s were local to my environment. Sorry for the noise.


tariqksoliman (changes requested):
Hi @Raydman and thanks for the PR! The following's me co-jointing reviewing with AI:

- Bug — indicator gets stuck. queryDataValue only calls back on HTTP 200 with values. On nodata/off-raster/errors nothing fires, pendingValueReads never reaches 0, and highlightInLegend() is never called — legendMatches is emptied but the Legend keeps the previous marker. Repro: hover DEM, then move over the bay; readout blanks, marker stays. Needs a failure path (or finally) so the legend is republished.
- Stale async writes. pendingValueReads is per-idPixel call and legendMatches is only reset on a new hover, so a slow callback from an earlier hover can write into the current set and re-highlight. Suggest a hover sequence id and drop callbacks whose id != current.
- Try mix-blend-mode: difference on the ramp marker (e.g. white bar) so it inverts against whatever color is beneath it; the current --color-mmgis + 1px outline disappears on parts of the ramp. Consider dropping the 120 ms transition — it adds perceived lag. Maybe a little arrow indicator could be better. Really we shouldn't use a blue indicator line since, in cases where the colorramp is blue too, there's too little contrast to see it. 
- The indicator in the LegendTool and the value in the IdentifierTool don't always seem to align as I'd think, which could be related to the first two bullet points.
- Nearest-number fallback on discrete rows can light the wrong row (hideFromLegend rows aren't rendered; IDs 1, 2, 50 snap 27 → 50). Keep nearest-match for pushScale ramps, require exact match for rows.
   - I didn't test this one (AI found only) but it reminds me that I need to add a discrete colorramp layer to the reference-mission.
- Remove the design-rationale comments (the legendMatches/pendingValueReads block, highlightEntries header, data-legend-* attr comments, marker css comments) — repo style is 1–2 lines; move the reasoning into the PR description.
- Add tests — both tools already have tests/*.spec.js; a hover → row css / marker exists → mouseout → cleared check is cheap.

Re "test suite is broken on development": I don't think so — createReferenceMission derives the name from the variant, so { setupReferenceMission: true } with no mission is the intended call (tests/global-setup.js does exactly this). 404s are most likely missing admin session in that environment.

Thanks!

Raydman:
Thanks @tariqksoliman — the review was thorough, and the spelled-out repros made these quick to confirm.

Indicator gets stuck. Fixed. Every route out of queryDataValue now reports exactly once, a value or null, so a nodata, off-raster or errored read settles the pending count and the legend is republished. The once-only guard also closes a pre-existing path where a string payload could invoke the callback twice.

Stale async writes. Fixed with the hover sequence id you suggested. Each read captures the layer it was issued for and the hover it belongs to, and drops itself if that hover is over.

Marker contrast. The indicator is now an arrow to the right of the bar pointing at it, mix-blend-mode: difference and no transition. The blend matters more there than on a bar: it crosses both the dark panel and the white numerals it overlaps, and inverts against each.

Indicator and value not aligning. You were right that this was real, but it turned out to be its own defect rather than a symptom of the first two. The mark was placed at the nearest stop, so it quantised — two readings in one band landed on the same pixel. It was also anchored to the label text, which sits at (i+0.5)/n, while the ticks run i/(n-1) across the full bar; the ticks are the true scale, since on a linear 0–800 ramp 73.55 m belongs at (800−73.55)/800 = 90.8%, which is the tick geometry exactly. Re-anchored to the ticks, every stop now lands on its own tick mark.

Nearest-match on discrete rows. Fixed as specified: nearest is ramps only, rows need an exact match. Your 1, 2, 50 case with a reading of 27 lights nothing, with a unit test pinning it. One judgment call I'd like your read on — a row label naming a span like 10 - 20 still matches a reading inside it. That's containment rather than snapping, so it doesn't reintroduce the bug, and it's what makes a binned legend usable. Easy to drop if you'd rather rows were exact and nothing else.

Comments. Trimmed to one or two lines throughout; the reasoning is in the PR description.

Tests. 34 unit cases and 10 E2E. The resolution logic lives in legendHighlight.js, which imports nothing from src/essence, so it unit tests directly. The E2E builds its ramp through the public overwriteLegends API rather than depending on mission data — your discrete colorramp note would let it use a real layer instead, and I'd be glad to follow up with a PR against the blueprint.

On the test suite. You're right and I was wrong. createReferenceMission derives the name from the variant, and global-setup does create an admin, log in and send the cookie. My 404s were local to my environment, and I've corrected the PR description.

One note for CI: the E2E specs run against the production bundle, so they need npm run build ahead of them. Mine passed against a stale bundle until I rebuilt — which is incidentally how I confirmed the new tests actually detect these changes.

image

PR #1093, opened 2026-09-20
#1092 Fix hanging webhook entries request on empty configuration

### LINKED ISSUE #1092: [Bug]: Webhook entries request hangs when no configuration exists
### Checked for duplicates

Yes - I've already checked.

### Describe the bug

On a fresh installation, `GET /api/webhooks/entries` never completes when the `webhooks` table contains no rows. The route only calls `res.send(...)` inside `if (sets && sets.length > 0)`, so an empty array leaves the request open until the client or proxy times out.

This also affects the Configure > Webhooks page, which calls this endpoint when it mounts.

### What did you expect?

The endpoint should return a successful response with an empty entries array:

```json
{
  "status": "success",
  "body": {
    "entries": []
  }
}
```

### Reproducible steps

```bash
1. Start MMGIS with a fresh database so the webhooks table has no rows.
2. Sign in as an administrator.
3. Request GET /api/webhooks/entries, or open Configure > Webhooks.
4. Observe that the request remains pending instead of returning an empty list.
```

### Environment

```bash
- MMGIS development branch
- Version: 5.4.1-20260916
- Reproduced by tracing the empty-result path in plugins/core/backend/Webhooks/routes/webhookutils.js
```

I am happy to submit a small fix with regression coverage.

Benziza:
## Purpose
- Ensure `GET /api/webhooks/entries` completes when no webhook configuration exists.
- Preserve the existing response shape for stored webhook configurations.

## Proposed Changes
- [FIX] Always send a successful entries response, including `entries: []` for an empty table.
- [ADD] Extract response formatting into a small helper that can be tested without a database connection.
- [ADD] Cover both empty and populated webhook entry responses with unit tests.

## Issues
- Closes #1092

## Testing
- `PLAYWRIGHT_TEST_UNIT_ONLY=true npx playwright test tests/unit/webhookEntries.spec.js --project=chromium --workers=1` — 2 passed.
- `NODE_ENV=test npx eslint plugins/core/backend/Webhooks/routes/webhookutils.js plugins/core/backend/Webhooks/routes/buildEntriesResponse.js tests/unit/webhookEntries.spec.js` — passed.


tariqksoliman (approved):
Good catch and LGTM, thanks!

Benziza:
Thanks for the review and approval Tariq! Appreciate it.

PR #1095, opened 2026-09-22
ci: add DockSec container scan with SARIF upload

advaitpatel:
## Purpose

Adds a GitHub Actions workflow that scans the `Dockerfile` with [DockSec](https://github.com/OWASP/DockSec)
and uploads the results to the Security tab as SARIF, so container findings are
annotated inline on pull requests. DockSec is an OWASP Lab Project (MIT); it
runs Trivy, Hadolint, and Docker Scout locally, then ranks every finding by
severity plus [EPSS](https://www.first.org/epss/) exploitation likelihood and
correlates them into attack chains, instead of handing back a flat list of
raw findings. This workflow uses `scan_only` mode, which reports the
Dockerfile misconfiguration and CVE findings through SARIF without the
AI-assisted prioritization/correlation pass.

## Proposed Changes

- [ADD] `.github/workflows/docksec.yml` - the only file this PR touches.

## Issues

None linked. This is a small, additive CI change; happy to open an issue first
if you would rather discuss the approach that way.

## Testing

DockSec scores the current `Dockerfile` at **70/100**, with 8 findings, across
the multi-stage `oraclelinux:8.9` build:

```
DL3041 warning  lines 9, 112  Specify version with dnf install -y <package>-<version>
DL3040 warning  line 33       `dnf clean all` missing after dnf command
DL3046 warning  line 125      useradd without flag -l and high UID will result in
                               excessively large Image
DL4006 warning  line 40       Set SHELL -o pipefail before RUN with a pipe in it
```

`DL3046` is the one specific to this file: an unflagged `useradd` with a high
UID writes a sparse `/var/log/lastlog` sized to that UID, which on some
filesystems and backup tools turns into real disk and I/O cost per layer - a
correctness issue that happens to show up as a security-adjacent Dockerfile
finding.

Verification performed before opening this:

- Ran DockSec against this repository's real Dockerfile locally
  (`docksec Dockerfile --scan-only`), which produced the score and findings
  above.
- Validated the workflow file with `actionlint` - no warnings.
- Verified the DockSec Action's SARIF output end-to-end against the published
  container (`ghcr.io/owasp/docksec:latest`): it writes a valid SARIF 2.1.0
  report with relative `Dockerfile` URIs into the workspace, which is what
  `upload-sarif` reads.

What this workflow does not do: it runs `scan_only`, so no API key and no AI
call; it sets no `fail_on`, so it reports without gating merges; and it scans
the Dockerfile only, not a built image, so the CVE layer of the base image is
not covered here. All three actions are pinned to a commit SHA,
`persist-credentials: false` is set on checkout, and the SARIF upload is
skipped on fork pull requests, where `security-events: write` does not apply
and the upload would otherwise fail as a red required check.

Triggered on the `development` branch, to match how this repository develops.

---

Disclosure: I used an AI assistant (Claude) to draft this description and
generate the workflow file from DockSec's documented Action inputs. The scan
result quoted above is from a run I did myself.

Disclosure of interest: I am the project lead for DockSec, the tool this
workflow adds. Happy to close this if it is not a fit - no hard feelings either
way.


advaitpatel:
Hi @tariqksoliman can I please get your review here? Happy to share more about the tool if needed. Thanks!

